PR #57, opened 2018-04-05
Create RSKIP57 about Derivation Path for Hierarchical Deterministic Wallets

RSKilan:


SergioDemianLerner:
Pull request looks good. However I think the RSKIP name should be "Derivation Path for Hierarchical Deterministic Wallets", because "Derivation Path" alone is unclear.

SergioDemianLerner:
All BIP URL paths read "bipc" when it should read "bips".

PR #60, opened 2018-04-25
RSKIP Checksum Address Encoding

ilanolkies:
RSKIP-0060 **describes an address checksum** mechanism that can be implemented in any network, to prevent the misuse of these and avoid typing errors

### DIFF
--- a/IPs/RSKIP60.md
+++ b/IPs/RSKIP60.md
@@ -4,7 +4,7 @@ RSKIP:
 	Title: Checksum Address Encoding
 	Authors:
 		Julian Len <julianlen@gmail.com>
-		Ilan Olkies <ilanolkies@gmail.com>
+		Ilan Olkies <ilanolkies@outlook.com>
 	Type: Standard Track
 	Created: 2018-25-06
 ```
--- a/IPs/RSKIP60.md
+++ b/IPs/RSKIP60.md
@@ -1,19 +1,16 @@
-# Checksum address encoding
+# Checksum Address Encoding
 ```
 RSKIP:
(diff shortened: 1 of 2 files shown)


IPs/RSKIP60.md
```diff
@@ -0,0 +1,89 @@
...
+## Abstract
+
+Addresses can be validated using an injective function that makes capital letters redundant.
+RSKIP-0060 **describes an address checksum mechanism** that can be implemented in any network based on [EIP-55](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md).
+
+## Specification
+```python
+def checksum_encode(addr, chain_id = None):
+    o = '0x'
+    addr_prefix = '' if None else str(chain_id) + addr.lower()
```
juli: This line is wrong, if addr is `b''` then it makes no sense to do `.lower` and if it is a string then it makes no sense to to `.hex()` in the following lines. `if None` is constant.

IPs/RSKIP60.md
```diff
@@ -0,0 +1,89 @@
...
+- Avoid typing confusion in adresses.
+- Differentiate addresses of different networks.
+
+## Abstract
+
+Addresses can be validated using an injective function that makes capital letters redundant.
+RSKIP-0060 **describes an address checksum mechanism** that can be implemented in any network based on [EIP-55](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md).
+
+## Specification
+```python
```
SergioDemianLerner: What is this, Python3 or Python2 ?

PR #71, opened 2018-05-28
RSKIP60 - Change seprator to 0x

ilanolkies:


IPs/RSKIP60.md
```diff
@@ -23,7 +24,7 @@ In Javascript:
 ```javascript
 function toChecksumAddress(address, chainId = null) {
     const strip_address = stripHexPrefix(address).toLowerCase()
-    const prefix = chainId != null ? (chainId.toString() + '|') : ''
+    const prefix = chainId != null ? (chainId.toString() + '0x') : ''
```
SergioDemianLerner: Why? There is no justification in the PR about this change. Also if you look up what gets hashed, it just mixes the numerical representation of chainId with the "0" of the hex prefix. It would look like "600x00......". Does this looks clear to you? Maybe adding both separators could be clearer "60|0x00..."

SergioDemianLerner:
Making compatibility-breaking changes to address checksums which the community needs asap doesn't look like a good idea to me.

IPs/RSKIP60.md
aeidelman: I agree with @SergioDemianLerner comment about describing and justifying changes in the PR itself, without it we should not even pay attention to the proposal. Having said that, the choice of delimiter is now a matter of personal preferences and I don't think it's a good idea to have this blocked for this reason. I strongly recommend to move forward with this change, taking into consideration that the implementations already done are considering 0x as delimiter.

SergioDemianLerner:
Ok. I'll approve, is up to the community (wallet builders) to decide and use it.

PR #74, opened 2018-06-10
TWPRO

martinmedina:
Draft ready for peer review

### DIFF
--- a/IPs/RSKIP-TWPRO.md
+++ b/IPs/RSKIP-TWPRO.md
@@ -1,0 +1,62 @@
+# Trustless Work Provider Protocol (TWPRO)
+
+|RSKIP          |           |
+| :------------ |:-------------|
+|**Title**      | Trustless Work Provider Protocol (TWPRO) |
+|**Created**    |06-JUL-18 |
+|**Author**     |JR, MM |
+|**Purpose**    |Usa, Sec |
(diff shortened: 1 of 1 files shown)


SergioDemianLerner:
This could be a backup mechanism, but by not means it should be the only node the mining pool uses. If that would be the case, it would defeat the propose of mining as a decentralization tool. The TWPRO node would be a defecto central party that can attack the network. I suggest that, in case the full node maintained by the mining pool is temporarily down, a local client should contact several TWPROs, and query each one to find a common parent (e.g. majority voting), then ask one of them to create a block for that parent. All this must happen quickly enough so the parent doesn't change.
In fact, this is like creating a UNL consensus fallback system (see Ripple).

martinmedina:
Thanks for the comments! 

I do like the idea of querying other nodes for the blockchain tips, etc, etc. These nodes do not necessary need to be TWPRO nodes, don't they ?
Because methods for retrieving that information are available on any RSK full node.

I'll add this to the RSKIP with a strong suggestion for miners to connect to one TWPRO node and also to other nodes for blockchain tip validation. However, mining pools should be able to work even if only a connection to a TWPRO is provided. 
Implementation wide, an scenario with only a connection to a TWPRO looks like a good MVP for mining pools. Later, blockchain tip validation can be added on a next version.

PR #98, opened 2018-08-24
Disable fallback mining in advance

aeidelman:


SergioDemianLerner:
Ping. Will be there an RSKIP related to this?

PR #106, opened 2019-02-08
RFS-298 - New BTO utils precompiled contract

amendelzon:
This RSKIP proposes a new precompiled contract with a set of four functions.


IPs/RSKIP106.md
```diff
@@ -0,0 +1,135 @@
...
+|**Complexity** |1 |
+|**Status**     |Draft |
+
+## Abstract
+
+A new precompiled contract (i.e., native, hardwired onto the RSK consensus layer) is introduced. This contract contains a few complex BTO-related operations that help leverage some of the work BTO contracts need to repeatedly perform.
+
+## Motivation
+
+With the arising need of BTOs (ICOs over Bitcoin) on top of the RSK network, a need for different BTO token and/or token-emitting contracts to perform certain complex BTO-related operations on a regular basis is in place. If written in solidity (or assembly WOLOG) and executed on the RVM, each of these functions would be expensive - and even more expensive to write and test every time. Furthermore, different implementations can lead to errors potentially protruding to loss of funds. A well-written, tested and precompiled function set that covers these operations will leverage the complexity on the BTO contracts' side while at the same time enforcing a certain level of security by implementing these as part of the RSK consensus.
```
SergioDemianLerner: Say "Bitcoin Token Offerings"

SergioDemianLerner: Instead of ICOs

SergioDemianLerner: What do you mean with "test every time" ? It it's a library then it can be tested just once.

IPs/RSKIP106.md
```diff
@@ -0,0 +1,135 @@
...
+
+Given the base58Check-encoded extended public key `tpubDCwK6XsmwUx641qZ6Uyb2pcZXMCoFcyNBFZLbxwJT4iwgmUhY9DsW9FgaFUNE5s1KtM1upxpDuCvx1TbAqQMLMGFF5b1F6KpRpuHcLwcp4k` we would have that:
+
+```
+extractPublicKeyFromExtendedPublicKey('tpubDCwK6XsmwUx641qZ6Uyb2pcZXMCoFcyNBFZLbxwJT4iwgmUhY9DsW9FgaFUNE5s1KtM1upxpDuCvx1TbAqQMLMGFF5b1F6KpRpuHcLwcp4k') => '0x0360a7239100410a4f6019778fce803c6b168a9b1922a0eafbdff11e866070d9f8'
+```
+
+### getMultisigScriptHash
+
+The method `getMultisigScriptHash(uint8 minimumSignatures, bytes[] publicKeys) returns (bytes)` takes as input a minimum required number of signatures and an array of (either compressed or uncompressed) secp256k1 public keys and returns the 20-byte hash (i.e., output of `hash160`) of the Bitcoin N-of-M multisig scriptPub corresponding to the given arguments. It is important to notice that the compressed version of the given public keys will be used to generate the script. Therefore, public key compression will not make a difference in the function output.
```
SergioDemianLerner: How are you planning to inject the "nonce" to generate different addresses for each contributor? Is this done by generating new pubkeys with increasing sequential numbers for BIP32?

IPs/RSKIP106.md
amendelzon: :+1:

amendelzon: I was thinking about different people writing their own BTO contracts and potentially using different libraries for this, or even writing their own. It is true, if you reused an existing solidity library for this, you wouldn't have to retest things. That said, I think that model doesn't actually enforce that reuse, but rather makes it an optional choice within the development process.

IPs/RSKIP106.md
amendelzon: Exactly. Essentially each public key within `publicKeys` is itself a derived public key from a given master key (or another derived key, without loss of generality). That would be the case if implementing, for example, [BIP45](https://github.com/bitcoin/bips/blob/master/bip-0045.mediawiki). But the thing here is we are leaving the freedom of choice to the actual BTO implementer.

IPs/RSKIP106.md
```diff
@@ -0,0 +1,135 @@
...
+A new precompiled contract is to be accessible at the `0x0000000000000000000000000000000000000009` address. It will implement the following functions (signatures and return values are as follows):
+
+- `toBase58Check(bytes, uint8) returns (string)`
+- `deriveExtendedPublicKey(string xpub, string path) returns (string)`
+- `extractPublicKeyFromExtendedPublicKey(string xpub) returns (bytes)`
+- `getMultisigScriptHash(uint8 minimumSignatures, bytes[]) returns (bytes)`
+
+## Error handling
+
+In case of error, each of these methods behaves as if a solidity `assert` statement was being used. That is, contract state is reverted and all the gas is consumed.
```
SergioDemianLerner: Are you sure this works in your reference implementation? Because calling precompiles from external transactions currently does not support REVERT. You should either support REVERT or prevent these functions from being called from exteernal transactions

IPs/RSKIP106.md
```diff
@@ -0,0 +1,135 @@
...
+
+```
+getMultisigScriptHash(2, ['0x02276a07b202503d39a43896300224fb649818d1486b0eefde6c9cd7cf1e32eed8', '0x0363e3de0d55459387f221dcda80f9283e303a08bcfeb9cf32875a43fefc7ecb1f', '0x03173019874589385b0295b839bb70e5d3c33178c110bacd872e245b656b3e8f43']) => '0xafa22d52922aa6657417c55b6cebdf708b594576'
+```
+
+_Note_: The public keys used here are compressed, but might as well be uncompressed. The output should be exactly the same since the function handles the conversion.
+
+## Rationale
+
+TODO (if any)
```
SergioDemianLerner: Add the gas consumption for each method somewhere.

amendelzon: Perfect. Will do when we get the final estimations in.

amendelzon: I've added in the gas consumptions.

PR #107, opened 2019-02-17
Node storage compression RSKIP

SergioDemianLerner:


tinchou:
I read "if lvalue>0 and hasLongVal" but `lvalue` isn't part of the specification.

Also, don't forget to update this with the new idea to encode the shared path!

tinchou:
Do you feel we really need to include `leftTreeSize`?

PR #116, opened 2019-03-13
Failure of SSTORE on Low-Gas Recursive CALLs

SergioDemianLerner:


SergioDemianLerner:
Hi @ajlopezrsk . For some reason, the PR was merged before your comments were taken into consideration. I will address them now and modify the RSKIP accordingly.

1. Links will be added about the danger of recursive calls. For example, https://medium.com/swlh/the-story-of-the-dao-its-history-and-consequences-71e6a8a551ee . Also the EIP1283 would have introduced a potential recursive-call vulnerability. This EIP can enable EIP1283 without preventing the vulnerability.

2. After some discussions, the magic number 8000 has been lowered to 6400, which comes from the cost of a call, which is 700, plus the cost of a call-back, which is 700, plus the minimum cost of a SSTORE, which is 5000. Therefore a contract passing less than 6400 would never allow recursive executions of SSTORE

3. In the example you give, the SSTORE does not fail, because the lock is decremented on return

PR #120, opened 2019-04-23
Shifting Opcodes

smcsicardi:
Created the RSKIP to implement the new opcodes of Shift Left, Shift Right and Arithmetic Shift Right


SergioDemianLerner:
This RSKIP replaces old RSKIP05

PR #125, opened 2019-05-10
Create2

smcsicardi:


chompomonim:
There is another proposal called `CREATECHILD` which is competing with `CREATE2`. Should RSK have both of them? :/

https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP59.md

smcsicardi:
> There is another proposal called `CREATECHILD` which is competing with `CREATE2`. Should RSK have both of them? :/
> 
> https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP59.md

Hi there! This instruction is mainly to maintain compatibility with the Ethereum VM, so contracts can be ran easily without many changes in both blockhains. Also, although the proposal was accepted the opcode was not implemented.

SergioDemianLerner:
RSKIP59 (older) was rejected in RSK BECAUSE CREATE2 was accepted in Ethereum. Compatibility was higher priority.

IPs/RSKIP125.md
```diff
@@ -0,0 +1,106 @@
...
+|**Complexity** |1 |
+|**Status**     |Draft |
+
+## Abstract
+
+The purpose of this IP is to implement the CREATE2 opcode, as added to the Ethereum VM in the [EIP 1014](http://eips.ethereum.org/EIPS/eip-1014)
+
+## Motivation
+
+Allows interactions to (actually or counterfactually in channels) be made with addresses that do not exist yet on-chain but can be relied on to only possibly eventually contain code that has been created by a particular piece of init code. Important for state-channel use cases that involve counterfactual interactions with contracts.
```
SergioDemianLerner: "Allow interactions" statement is confusing. Consider the shorter "Enables the creation of contracts with deterministic and repeatable addresses. This functionality is required for state channels having counterfactual interactions with a contract."

smcsicardi: This is the original motivation text from the EIP, I agree with you, it is confusing, I can make the change if you want anyways

PR #141, opened 2019-09-27
Wasabi+1 network upgrade proposal

diega:


IPs/RSKIP141.md
```diff
@@ -0,0 +1,59 @@
...
+- Add EXTCODEHASH opcode
+- Add CHAINID opcode (ref. [EIP-1344](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1344.md))
+- 2WP Locking Cap
+- 2WP Onchain Events
+- ValidateEthPow precompiled contract
+
+### Tentatively Accepted EIPs
+
+- ECADD, ECMUL and Pairing check precompiled contract
+- SELF_BALANCE opcode and SLOAD, BALANCE and EXTCODEHASH cost reduction (ref. [EIP-1884](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1884.md))
```
SergioDemianLerner: Most are cost increases rather than cost reductions. only SELF_BALANCE is a reduction.

IPs/RSKIP141.md
```diff
@@ -0,0 +1,59 @@
...
+- 2WP Segwit Compatibility
+- Disable TXINDEX, DUPN, SWAPN, OP_HEADER opcodes
+- CalculateMMR precompiled contract
```
SergioDemianLerner: As I remember, having a precompiled contract for this was an option that was discarded. The MMR would be computed by native code, right?

IPs/RSKIP141.md
```diff
@@ -0,0 +1,59 @@
...
+### Proposed EIPs
+
+- Decentralize feePerKb calculation for precompiled Bridge contract
+- Costs reductions (ref. [EIP-1108](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1108.md) / [EIP-2028](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2028.md) / [EIP-2200](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2200.md))
+- Add Blake2 compression function F precompile (ref. [EIP-152](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md))
+- GetRefund precompiled contract (ref. [RSKIP-139](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP139.md))
+- Add internal transactions hash in block header
+- Add bloom filter hash in block header
+- Precompiled Bridge contract events in Solidity format
+
```
SergioDemianLerner: We also mentioned replacing Bloom value with hash(bloom) when computing the transaction id (all the RPC interface remains equal). The objective is to reduce FlyClient proofs (bloom takes 256 bytes, which are totally unnecessary) and eventually reduce storage for nodes that do not care about bloom

IPs/RSKIP141.md
```diff
@@ -0,0 +1,59 @@
...
+### Included RSKIPs
+
+- TBD
+
+### Accepted RSKIPs
+
+- Add EXTCODEHASH opcode
+- Add CHAINID opcode (ref. [EIP-1344](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1344.md))
+- 2WP Locking Cap
```
josedahlquist:
```suggestion
- [2WP Locking Cap](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP134.md)
```

PR #160, opened 2020-02-27
docs: Change Testnet coin type 37310 --> 1 

bguiz:
## What

- In the derivation path, the coin type for Testnet was `37310` and is now `1`

## Why

- [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)
  has been updated to make *all* Testnets use `1` as the coin type,
  and thus this RSKIP is out of date


alepc253:
We can't change that... many wallets are using that number in the derivation path

bguiz:
> We can't change that... many wallets are using that number in the derivation path

@alepc253 OK sure, closing this PR.

PR #178, opened 2020-10-14
External Confirmation Hashrate

SergioDemianLerner:


IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
...The only contribution to the cryptoeconomic security of an independent proof-of-work chain without subsidy is the existence of high transaction fees. In the case of sidechain 1:1 pegged to a primary chain that is protected by merge-mining, there is a high strategic alignment between the primary chain miners and merge-miners, so the honest majority and the fact that attacks to the secondary chain often become attributable are the two main assurances for the security for the sidechain. Nevertheless, RSK went further, and deployed the Armadillo system to provide fork-awareness. This subsystem creates a time window to trigger an emergency response by the community, in case of long reorg attempts.  Nevertheless, Armadillo requires a separate communication channel and the existence of at least one honest node monitoring the Bitcoin network. The SFAMM protocol and security model captures the benefits of Armadillo, makes Armadillo decentralized, and formalizes the protocol algorithms, assumptions and guarantees. The SFAMM protocol is ideal for RSK as it focuses not only on the hashrate capability of the attacker but also on the processing time that will be required for a successful attack.
```
salindne, line 27: The only contribution to the cryptoeconomic security of an independent proof-of-work chain without subsidy is the existence of high transaction fees. In the case of a side-chain 1:1 pegged to a primary chain that is protected by merge-mining, there is a strategic alignment between the primary chain miners and merge-miners that results in two main assurances for the security of the side-chain:  the honest majority holds and the fact that attacks to the secondary chain often become attributable. Nevertheless, RSK has gone further and deployed the Armadillo system to provide fork-awareness. This subsystem creates a time window to trigger an emergency response by the community, in case of long reorg attempts.  Nevertheless, Armadillo requires a separate communication channel and the existence of at least one honest node monitoring the Bitcoin network. The SFAMM protocol and security model captures the benefits of Armadillo, makes Armadillo decentralized, and formalizes the protocol algorithms, assumptions and guarantees. The SFAMM protocol is ideal for RSK as it focuses not only on the hashrate capability of the attacker but also on the processing time that will be required for a successful attack.

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+
+This RSKIP proposes a consensus change that enables RSK nodes to:
+
+- compute the active SHA256 hashrate and detect variations on the active hashrate that are compatible with an ongoing attack
+
+- Accumulate all the active SHA256 hashrate in the RSK blockchain, with independence of the amount of hashrate that is directly involved in merge-mining. This hashrate means more accumulated work, increasing the security against certain double-spend attacks. Also, the hashrate accumulated not only includes 100% of Bitcoin hashrate, but also hashrate from bitcoin forks.
```
salindne, line 33: - Accumulate all the active SHA256 hashrate in the RSK blockchain, with independence of the amount of hashrate that is directly involved in merge-mining. This accumulated SHA256 hashrate results in more accumulated work, increasing the security against certain double-spend attacks. Also, the hashrate accumulated not only includes 100% of Bitcoin hashrate, but also hashrate from bitcoin forks.

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+
+The block in encoded as:
+
+```
+blockHeader, transactions, uncles
+```
+
+And *uncles* is a list of RSK block headers. We define a **heavyblock** to be a *mergeMiningProof* with certain properties that will be verified in consensus.
+
+We extend the list uncles to include, after all RSK block headers, a number of heavyblock elements (defined later). To maintain semantic coherence, the *uncles* field is renamed *difficultyContributors*. The field *uncleCount* is renamed as *difficultyContribution*.
```
salindne, line 64: We extend the list of uncles to include, after all RSK block headers, a number of heavyblock elements (defined later). To maintain semantic coherence, the *uncles* field is renamed *difficultyContributors*. The field *uncleCount* is renamed as *difficultyContribution*.

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+| Bits           | Current [target](https://en.bitcoin.it/wiki/Target) in compact format | 4            |
+| Nonce          | 32-bit number                                                | 4            |
+
+We define how to convert the Bits field into a target difficulty that can be compared with the RSK block difficulty.
+
+The Bits field is stored in a compact format. Let Bits be a byte array *{ E,Q1,Q2,Q3 }*. Let M be the integer value *(Q1 << 16 + Q2 << 8 + Q3)* (that is the *Q* array interpreted in big-endian). The Bitcoin target is computed as *M\*2^(8(E-3))*. We define the *btcDifficulty0 = 2^256 / target*. Note that the *btcDifficulty0* is *2^32* times higher than the Bitcoin block difficulty.
+
+### Validation
+
+We'll define the function *getPoWInfo()* to check the validity of all kinds of blocks (mainchain, uncles and elements in the *heavyblocks* list). This means that the function that checks the validity of the proof of work of normal RSK blocks is changed to call getPoWInfo(). This is because some of the RSK blocks can be anchor heavyblocks.
```
salindne, line 99: We define the function *getPoWInfo()* to check the validity of all kinds of blocks (mainchain, uncles and elements in the *heavyblocks* list). This means that the function that checks the validity of the proof of work of normal RSK blocks is changed to call getPoWInfo(), since some of the RSK blocks can be anchor heavyblocks.

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+
+But before, we'll define an auxiliary function *getMMInfo()*:
```
salindne, line 101: First we define an auxiliary function *getMMInfo()*:

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+
+Note that the midstate of *SHA256Digest* class in rskj uses a different format that the one used in this description. We call our variable midstate40 to distinguish between them.
+
+Also it is important to note that the above algorithm requires that for a block that doesn't have the merge-mining the tail size must be equal or grater than 169 bytes, but less than 233 bytes or the midstate is in fact the SHA256 initial state, while for a header that does have the merge mining tag, the tail size must be less than 169 bytes. This last check is the same as how the current check *ProofOfWorkRule* performs as:
```
salindne, line 163: Also it is important to note that the above algorithm requires that for a block that doesn't have the merge-mining the tail size must be equal or greater than 169 bytes, but less than 233 bytes or the midstate is in fact the SHA256 initial state, while for a header that does have the merge mining tag, the tail size must be less than 169 bytes. This last check is the same as how the current check *ProofOfWorkRule* performs as:

IPs/RSKIP178.md
```diff
@@ -0,0 +1,375 @@
...
+
+We propose the following functions:
+
+*A=btcDifficulty0/2*
+
+*R=difficulty/2*
+
+*C=btcDifficulty0*
+
+These formulas require using 256-bit arithmetic. This assignment overstates on average the RSK blockchain cumulative difficulty by at most 1/20 (the ratios of average block intervals between RSK and Bitcoin), which we consider too low to use the exact formulas.
```
salindne, line 276: These formulas require the use of 256-bit arithmetic. This assignment overstates on average the RSK blockchain cumulative difficulty by at most 1/20 (the ratios of average block intervals between RSK and Bitcoin), which we consider too low to use the exact formulas.

diegomasini (approved):
LGTM!

PR #186, opened 2020-11-20
RSKIP-??? - Active Federation creation block height

josedahlquist:


IPs/RSKIP186.md
```diff
@@ -0,0 +1,70 @@
...
+|**Layer**      |Core |
+|**Complexity** |1 |
+|**Status**     |Draft |
+
+## Abstract
+
+When the Federation change voting process finishes, the Bridge emmits a `commit_federation` event.
+This event can be used to provide a blockchain proof of the Federation address when requested. Nowadays doing this would imply navigating the blockchain to determine when it was emmited, or constantly monitoring said events.
+
+This RSKIP proposes adding on chain information for easily fetch the block where the Federation was changed for the last time. With the block, a system could provide the proof that it belongs to the mainchain.
```
marcos-iov, line 18:
```suggestion
This RSKIP proposes adding on-chain information to easily fetch the block where the Federation was changed for the last time. With the block, a system could provide proof that it belongs to the mainchain.
```

IPs/RSKIP186.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Motivation
+
+When a user gets the Federation address they usually interact with a node/service and have to trust on this information. This exposes a risk where a man-in-the-middle kind of attack could lead users to send their funds to invalid address thus losing them.
```
marcos-iov, line 22:
```suggestion
When a user gets the Federation address they usually interact with a node/service and have to trust this information. This exposes a risk where a man-in-the-middle kind of attack could lead users to send their funds to an invalid address thus losing them.
```

IPs/RSKIP186.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Specification
+
+### Storage changes
+
+The Bridge will store two new values:
+- activeFederationCreationBlockHeight. (long)
+  - Defaults to the existing Federation activation height 
+- nextFederationCreationBlockHeight. (long)
+  - Defaults to the existing Federation activation height
```
marcos-iov, line 32: The names sound a bit confusing to me, I had to read a couple of times what the difference between each is.

Maybe:
```
federationCreationBlockHeight
federationActivationBlockHeight
```

What do you think?

josedahlquist: - activeFederationCreationBlockHeight: stores the active Federation creation height. This value is the one the user would get
- nextFederationCreationBlockHeight: this is a temporary value created when a new Federation is committed. This committed Federation won't be activated until the network's configuration delay is achieved.

I'll try better explaining this in the RSKIP.

Thanks!

marcos-iov (approved):

donequis:
This doesn't seem to solve the issue that is addressing at the description.

```When the Federation change voting process finishes, the Bridge emmits a `commit_federation` event.
This event can be used to provide a blockchain proof of the Federation address when requested. Nowadays doing this would imply navigating the blockchain to determine when it was emmited, or constantly monitoring said events.```

An user requesting the current federation can't validate with this event that the address provided it is actually the current federation. The reasonable thing would be to have a storage key with the federation address and present a storage proof of it or something similar.

josedahlquist:
> This doesn't seem to solve the issue that is addressing at the description.
> 
> `` When the Federation change voting process finishes, the Bridge emmits a `commit_federation` event. This event can be used to provide a blockchain proof of the Federation address when requested. Nowadays doing this would imply navigating the blockchain to determine when it was emmited, or constantly monitoring said events. ``
> 
> An user requesting the current federation can't validate with this event that the address provided it is actually the current federation. The reasonable thing would be to have a storage key with the federation address and present a storage proof of it or something similar.

You are correct @donequis, this is not enough to provide that functionality. But it's the required consensus change to be able to later create said functionality in the node itself (via an json-rpc method most likely)

donequis:
I think that this is not enough to create such functionality (at least not cleanly). With these changes a node will be able to find a block with the event and present a proof for such an event. The problem is that the user receiving the proof can't verify that the provided address is the actual federation address (i.e. there was no such an event after the given one). The user will be able to check that the address was at least active once in the past, which is not an enough guarantee. 

The only solution is to provide a proof of the height/tx execution and only then provide the event to the user. But it would be easier to save the federation address and send it directly to the user with a proof of storage

juli:
This can be solved offchain without a consensus change. I think this RSKIP does not make any senses.

PR #197, opened 2020-12-15
RSKIP197 -  Fix Precompile Calls Not Conforming With CALL Semantics

fedejinich:
Introducing a new mechanism to handle errors on failed precompiled calls


fedejinich:
closed in this one https://github.com/rsksmart/rskj/pull/1392

fedejinich:
I've just reopened this RSKIP with the specification @SergioDemianLerner @nagarev , any comments?

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+|**Created**    |15-12-2020 |
+|**Author**     |FJ |
+|**Purpose**    |USa |
+|**Layer**      |Core |
+|**Complexity** |2 |
+|**Status**     |Accepted |
+
+## Abstract
+
+This RSKIP specifies a new expected behavior for a precompiled contract call: any expected failed call pushes a zero into the stack and reverts the state before the execution.
```
SergioDemianLerner, line 15: What is an "expected failed call" ?

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Motivation
+
+Provides a mechanisim to handle errors for precompiled calls by checking the returned state. Before Iris hardfork the transaction wasn't reverted and consumed all the remaining gas from the entire call.
```
SergioDemianLerner, line 19: -> Before the iris hard-fork, a precompiled contract could not revert, but only rise an OOG, consuming all gas passed to the CALL.

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+
+Suppose that there is a Contract A that, somewhere, calls a Precompiled contract B with invalid data (or something that produces an expected failure), it's desirable to throw a kind "exception" that gives the chance to the caller (A) to handle and use it for it's own pouposes, instead of failing the entire execution.
+
+## Specification
+
+When `blockNumber >= IRIS_HF`, during a precompiled contract execution any expected failure should:
+1. Revert the execution state.
+2. Push a ZERO into the stack.
```
SergioDemianLerner, line 27: This specification is unclear. The ZERO or ONE is pushed in the stack according to the CALL semantics. Are you changing the CALL semantics? Are yo making a CALL to a pre-compiler respect this semantics ?

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+
+Lately this zero state can be used by the caller to decide what to do whenever it reaches the expected failed state.
+
```
SergioDemianLerner, line 30: unnecessary. This is already part of CALL spec.

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+Note: an expected failure should be considered by the precompiled itself, any other kind of runtime failures should still consume the entire provided gas.
+
```
SergioDemianLerner, line 32: unnecessary. Remove

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+        }
+        
+        // some logic
+
+        return true;
+    }
+}
+```
+
+This wouldn't have been posible, before introducing this RSKIP, because the entire execution would have failed whenever it get an error from the precompiled call.
```
SergioDemianLerner, line 74: This happens because you're passing all the gas(). If you had pass less amount of gas to the CALL, you could have checked the returned code.

Therefore the comment is incorrect.

I suggest you remove the comment, and the code.

SergioDemianLerner (changes requested):
As far as I understand this change, the RSKIPS should have a single line and nothing more:

"Add REVERT capability to pre-compiled contracts"

That's all that is needed to understand the change. There is no need for examples.

SergioDemianLerner:
I'm still trying to understand why this is a consensus change. The only reason I found is that this change might bring some unintended side-effects, so we synchronize the activation of this change with the Iris upgrade.

fedejinich:
> I'm still trying to understand why this is a consensus change. The only reason I found is that this change might bring some unintended side-effects, so we synchronize the activation of this change with the Iris upgrade.

Thats right, we are trying to synchronize some unexpected behavior from rskj that might affect a transaction final state

fedejinich:
here is a much simplified version of this RSKIP, no examples, a oneliner

IPs/RSKIP197.md
fedejinich: this part seems unclear to me also, I was trying to point those scenarios where you expect the precompiled to revert the transaction

IPs/RSKIP197.md
fedejinich: I've left this as motivation

IPs/RSKIP197.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Abstract
+
+This RSKIP specifies a new expected behavior for a precompiled contract call: any expected failed call pushes a zero into the stack and reverts the state before the execution.
+
+## Motivation
+
+Provides a mechanisim to handle errors for precompiled calls by checking the returned state. Before Iris hardfork the transaction wasn't reverted and consumed all the remaining gas from the entire call.
+
+Suppose that there is a Contract A that, somewhere, calls a Precompiled contract B with invalid data (or something that produces an expected failure), it's desirable to throw a kind "exception" that gives the chance to the caller (A) to handle and use it for it's own pouposes, instead of failing the entire execution.
```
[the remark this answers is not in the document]
fedejinich: removed

IPs/RSKIP197.md
fedejinich: removed

IPs/RSKIP197.md
fedejinich: removed

SergioDemianLerner (approved):

PR #213, opened 2021-02-02
Simple Transaction Versioning System

SergioDemianLerner:


IPs/RSKIP213.md
```diff
@@ -0,0 +1,77 @@
...
+|**Author**     |SDL|
+|**Purpose**    |Sca |
+|**Layer**      |Core |
+|**Complexity** |1|
+|**Status**     |Draft |
+
+
+# **Abstract**
+
+This RSKIP proposes the addition of a version field so that transaction parsing can start after transaction version has been verified. This reduces the amount of changes required for the node and software wallets.
```
salindne, line 17: This RSKIP proposes the addition of a version field so that transaction parsing can start after the transaction version has been verified. This reduces the amount of changes required for the node and software wallets.

IPs/RSKIP213.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Motivation
+
+Many features, like ephemeral calldata, multi-signed transactions and LTCP transaction compression require adding more information to transactions. However currently RSK transactions do not have a way to indicate a new transaction format is being used. We propose a transaction format that embeds a version field in the current nonce position, keeping high compatibility with wallets.
```
salindne, line 21: Many features, like ephemeral calldata, multi-signed transactions and LTCP transaction compression require adding more information to transactions. However currently RSK transactions do not have a way to indicate that a new transaction format is being used. We propose a transaction format that embeds a version field in the current nonce position, keeping high compatibility with wallets.

IPs/RSKIP213.md
```diff
@@ -0,0 +1,77 @@
...
+2. gasPrice
+3. gasLimit
+4. receiveAddress
+5. value
+6. data
+7. v
+8. r
+9. s
+
+The field nonce is renamed multiPurpose, and it contains an RLP list of two values: version and nonce. Version is encoded as an integer. The only version value accepted is 1. The conding must be minimal (no zero padding). Following RSKIPs can define additional versions. Transactions version 0 must use the previous format.  The RLP parser can detect if it is a transaction version 0 or version 1 by inspecting only the multiPurpose field (previously called nonce). If this element is a list (more precisely, starts with byte 0xc2), then the version is 1. If not, then it is the nonce, and the transaction version is 0.
```
salindne, line 56: The field nonce is renamed multiPurpose, and it contains an RLP list of two values: version and nonce. Version is encoded as an integer. The only value for version that is accepted is 1. The coding must be minimal (no zero padding). Following RSKIPs can define additional versions. Transactions version 0 must use the previous format.  The RLP parser can detect if it is a transaction version 0 or version 1 by inspecting only the multiPurpose field (previously called nonce). If this element is a list (more precisely, starts with byte 0xc2), then the version is 1. If not, then it is the nonce, and the transaction version is 0.

salindne (changes requested):

IPs/RSKIP213.md
SergioDemianLerner: Can you indicate the exact change required instead of repeating the paragraph?

IPs/RSKIP213.md
SergioDemianLerner: same here

IPs/RSKIP213.md
SergioDemianLerner: same here

IPs/RSKIP213.md
salindne: Add the word "the" in to the sentence "This RSKIP proposes the addition of a version field so that transaction parsing can start after **the** transaction version has been verified".

IPs/RSKIP213.md
salindne: The word that is added into "Many features, like ephemeral calldata, multi-signed transactions and LTCP transaction compression require adding more information to transactions. However currently RSK transactions do not have a way to indicate **that** a new transaction format is being used. We propose a transaction format that embeds a version field in the current nonce position, keeping high compatibility with wallets."

IPs/RSKIP213.md
salindne: The word "conding" -> coding

SergioDemianLerner: encoding is better

diegomasini (approved):
LGTM!

diegomasini (approved):
LGTM!

PR #215, opened 2021-02-04
Ephemeral Blockchain

SergioDemianLerner:


salindne (approved):
The following are some comments from slack channel:

**SM Comment:** 
> The logic to remove checkpoints is fine
> - check eligibility (enough epochs have passed)
> - check cumulative work: the cumulative work of the segments after the checkpoint must exceed that of the removal "candidate".


**SM Question:** 
> I wonder if there should be some minimum cumulative work to add a checkpoint. So weak  checkpoints are not added at all. I was going to suggest a cut off based on a moving average of weights .. but that won't work with large epochs... too few data points.

**SDL Response:** 
> Probably not needed because the epochs are assumed large and so all checkpoints should achieve high cumulative work. Having many short epochs is not desirable because too many states are then stored. 
> The important property for checkpoint removal is that it must be irreversible: even if the blockchain reorganizes, a removed checkpoint cannot be re-introduced.

**SM Question:** 
> Once removed, will there be a record of that epoch (that it use to be a checkpoint)?

PR #220, opened 2021-03-26
Obtain block or block header by hash or height in bridge methods

pamgonzalez:


marcos-iov (approved):
Add reference to RSKIP 199, since you are mentioning the block index

PR #221, opened 2021-03-31
Remove irrelevant statement.

SergioDemianLerner:
RSK removed those codes because they required maintaining a fork of Solidity, which it seems nobody wants to do.


smishraIOV (approved):
Even though the title says it all, it would be good to add a bit more detail in the abstract. The deleted line was was kind of informative.  

Can say "These opcodes are not part of EVM spec and are being deprecated on RSK"

PR #222, opened 2021-03-31
Restored removed entries

SergioDemianLerner:


SergioDemianLerner:
Some entries were removed by msitake.
The names were also sorted.

SergioDemianLerner:
It has already been corrected

PR #239, opened 2021-04-27
Reprice trie read

smishraIOV:
A proposal related to storage rent where we first reprice (increase the fixed cost of) some trie read op codes


diegomasini (approved):
LGTM!

PR #243, opened 2021-05-17
Intra-transaction Refunds

SergioDemianLerner:


IPs/RSKIP243.md
```diff
@@ -0,0 +1,77 @@
...
+If a contract call revers, the changes on *contracts_created* and *storage_cells_created* are not reverted. This implies that clearing a storage cell may trigger a refund even if the cell was already cleared. 
+
+The cap of the refund to 50% of the transaction gas consumed is removed.
+
+
+# Rationale
+
+Several related topics are analyzed in separate sections.
+
+### This proposal vs. [EIP2929](https://eips.ethereum.org/EIPS/eip-2929)
```
smishraIOV, line 46: Do you mean EIP-3298 (instead of 2929)? There are some connections to EIP-2929, but those are not direct. EIP-2929 was motivated primarily as DoS mitigation.

smishraIOV (approved):

IPs/RSKIP243.md
SergioDemianLerner: No. I'm really referring to the part of EIP2929 that proposes the access lists., not the part that changes the opcode costs. Access lists may at first sight look similar to creation lists, but serve a completely different purposes.
Access lists change cost to take into account intra-block caches in the node (read operations).
Creation lists change cost to take into account the case where nothing is written to SSD in the node. (write operations)

SergioDemianLerner: I'm pushing a new version with hopefully a more clear section.

SergioDemianLerner: @smishraIOV Tell me if the clarification is satisfactory for merging.

smishraIOV (approved):
Clarification regarding eip 2929 is fine. OK to merge

PR #184, opened 2020-11-19
RSKIP153 - Add BLAKE2 compression function `F` precompile

fedejinich:
This PR adds BLAKE2 F compression function, adopted first by ethereum and then by RSK, to maintain compatibility at EVM level. 

This RSKIP might be included in Iris hardfork.

PR #254, opened 2021-07-08
Update index with rskip153

fedejinich:
This PR updates the index to include the [RSKIP153](https://github.com/rsksmart/RSKIPs/pull/184)


fedejinich:
Since we are going to include it in IRIS HF, is it okey to tag it as `Accepted`? @SergioDemianLerner

PR #277, opened 2021-08-24
Changed Iris RSKIPs status

aeidelman:
Changed Iris RSKIPs status to Adopted


smishraIOV (approved):
Oh wow. Papyrus (141) was not marked "adopted" earlier! Good catch!

smishraIOV:
Usually, branches auto-delete on merge. I don't know why this patch needs explicit deletion. I'm leaving it as is.

PR #296, opened 2022-03-18
Update RSKIP144.md

javicid:


IPs/RSKIP144.md
```diff
@@ -19,47 +19,36 @@ RSK processes transactions from blocks one by one, in the specified order. This
...
-Therefore we suggest than once every k blocks (i.e. k=5) a block must be serial (a "serial block"), and the miner cannot specify any parallelism in the serial block. This will enable all postponed transactions that could not be parallelized to compete and be included in serial blocks. An alternative is to pseudo-randomize the choice of the serial block (i.e depending on the previous block hash). Adding this uncertanity however doesn't seem to provide any benefit.
+A transaction is connected to a partition if it is connected to any transaction in that partition. The miner maintains a writeMap and a readMap that store which partition/s write or read each storage key. The miner also maintains an accountMap that maps each sender account to a partition. The miner then executes transactions from the pool and keeps track of which storage keys the transaction reads and writes. After executing the transaction, the miner compares the transaction read and written keys with the readMap and writeMap. The following scenarios are possible:
```
SergioDemianLerner, line 38: Why does it need to keep an accountMap when the account is already a node in the trie ? 
By "storage key" do you mean contract storage keys or trie keys ?

IPs/RSKIP144.md
```diff
@@ -19,47 +19,36 @@ RSK processes transactions from blocks one by one, in the specified order. This
...
 
-Special treatment may be needed for recursive deletes, as used in suicides. 
+When a miner executes transactions to create an (unsolved) block, the miner must execute the transactions serially and decide which partition the transaction belongs to. The block gas limit is replaced by a per-partition gas limit. We say that a transaction is *connected* to another transaction if:
```
SergioDemianLerner, line 32: How the miner creates the partitions is not strictly a part of the consensus specification. It's part of the implementation. Therefore I suggest that all the description on how the miner creates the block is moved to another section: "Suggested Miner Implementation"

IPs/RSKIP144.md
javicid: You are right, an accountMap is not needed

IPs/RSKIP144.md
javicid: ok!

SergioDemianLerner (approved):

PR #300, opened 2022-05-17
New rskips site

max-barrera-2148:
Creating new site and setting up CD


_includes/indextable.html
```diff
@@ -25,6 +21,20 @@ <h2 id="{{status|slugify}}">{{status}}</h2>
           {% endif %}
         </thead>
         {% for page in rskips %}
+        {% if status == "Purpose and Guidelines" %}
+        <tr>
+          <td class="ipnum"><a href="{{page.url|relative_url}}">{{page.rskip|xml_escape}}</a></td>
+          {% if status == "Last Call" and page.last-call-deadline != undefined %}
```
patogallaiovlabs, line 27: This "if status == Last..." will never be true, bc is inside another "if status == Purpose...". Am i missing something?

IPs/RSKIP0.md
```diff
@@ -1,3 +1,15 @@
+---
+rskip: 0
+title: RSKIP Purpose and Guidelines
+description: A RSKIP is a design document providing information to the RSK community describing a new feature for RSK, its processes, or its environment.
+status: Purpose and Guidelines
+purpose: Sec
+author: JL (@julianlen)
+layer: Misc
```
patogallaiovlabs, line 8: Misc is not an option, at least I cannot see it in the list. Check "index.html", "Layer" section.

_includes/indextable.html
patogallaiovlabs: Also, I cannot see any diff between both blocks inside the if
if
 block1
else
 block2
 
 block1 and block2 have the same code.

index.html
```diff
@@ -0,0 +1,257 @@
...
+    <meta property="og:locale" content="en_US" />
+    <link rel="canonical" href="http://localhost:4000/" />
+    <meta property="og:url" content="http://localhost:4000/" />
+    <meta property="og:site_name" content="RSK Improvement Proposals" />
+    <meta name="twitter:card" content="summary" />
+    <meta name="twitter:site" content="@" />
+    <script type="application/ld+json">
+    {
+      "@type": "WebSite",
+      "url": "http://localhost:4000",
```
patogallaiovlabs, line 27: Could you change all places where "localhost:4000" appears and replace it for a template variable or something like that?

patogallaiovlabs (changes requested):
Apart from these small comments, I would suggest a Readme kind of file to document a step by step to start the server locally to test it.
I used only 2 commands, bundle install / bundle exec jekyll serve. But may not be trivial for people not used to this tech stack.

_includes/indextable.html
max-barrera-2148: Right, we're not using the status “last call” and there is some redundant code. I've cleaned it up and removed the unnecessary code. I've updated the Readme to include instructions on how to build the site locally to test. Thanks for the suggestion! I've already modified RSKIP0 and RSKIP0_TEMPLATE  to include the updated instructions on how to add the new IP using Jekyll. Basically, the main difference would be they would need to add the Jekyll front matter by adding three hyphens before and after the title, created date, author, etc.  Really the process is quite similar but if it's still not clear please let me know.

IPs/RSKIP0.md
max-barrera-2148: Hey, I left this as is for now, but I'm checking with Diego Masini if we'd need to update the "Layer" of this IP or update the list of "Layers" as soon as I know how we want to resolve this I will update it.

index.html
max-barrera-2148: This has been updated we're using Liquid syntax variable where applicable. Thank you.

patogallaiovlabs (approved):
lgtm 👍

IPs/RSKIP34.md
```diff
@@ -1 +1,50 @@
...

# **Specification**

New opcode: DATA

This opcode executes as STOP.

# **Copyright**

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
max-barrera-2148, line 1: I accidentally removed the Abstract, I need to add it back before merging the PR

PR #312, opened 2022-06-08
RSKIP144 update 2

ilanolkies:


IPs/RSKIP144.md
```diff
@@ -9,48 +9,88 @@
...
+An empty `partitionEnds` indicates that all transactions go in the _sequential partition_.
+
+Values in `partitionsEnd` must be greater than 0 and in ascending order. An empty `partitionEnds` indicates that all transactions go in the sequential partition. The maximum number of parallel partitions that the miner can specify is equal to the minimum number of cores required to run the RSK node, this maximum is called  `maxTransactionExecutionThreads`.
+
+## New block validation consensus
+
+The new consensus mechanism allows to reject blocks that, when executed in parallel, can result in different outputs.
 
-When a miner executes transactions to create an (unsolved) block, the miner must execute the transactions serially and decide which partition the transaction belongs to. The block gas limit is replaced by a per-partition gas limit. We say that a transaction is *connected* to another transaction if:
+For simplicity, three types of connections between transactions are identified:
```
javicid, line 52: Since "connections" are not something that actually exist between transactions, but something that we define to easily explain the algorithm, I would say that "We define a connection between two transactions if/when..."

IPs/RSKIP144.md
```diff
@@ -9,48 +9,88 @@
...
+- Both transactions are from the same sender account
+- Both transactions write the same storage key
```
javicid, line 54: These two are the same as Sergio pointed out in my PR. 

We can leave only the second one and then explain that transactions from the same sender are always connected because they modify the nonce.

IPs/RSKIP144.md
```diff
@@ -9,48 +9,88 @@
...
+- One transaction reads a key that the other transaction writes
 
-1. Both transactions are from the same sender account
-2. Both transactions write the same storage key
-3. One transaction reads the same storage key that the other transaction writes
+Any pair of transactions with any of these characteristics are required to be in the same _parallel partition_ to avoid non-deterministic results, or one needs to be executed in the _sequential partition_
```
javicid, line 57: Can be both in the sequential also. I would say that two connected transactions cannot be in different parallel partitions.

IPs/RSKIP144.md
```diff
@@ -9,48 +9,88 @@
...
+
+### Block validation algorithm
+
+A `readMap` and a `writeMap` is created per _parallel partition_. During execution of each partition,
+- when a storage key is read, it is marked in the `readMap`,
+- when a key is written, the key is marked in the `writeMap` map.
+
+When all parallel partitions have finished processing, the `readMap`s and `writeMap`s are scanned to find connections between transactions of different threads. This requires efficient maps that enable traversing the keys in ascending lexicographic order.
+- If a key belongs to a `writeMap` and a `readMap` of another _parallel partition_, then the block is considered invalid.
+- If a key belongs to a `writeMap` and a `writeMap` of another _parallel partition_, then the block is also considered invalid.
```
javicid, line 69: This ended up being different in the implementation. We created two maps readKeys and writtenKeys that map key -> thread/s. 

When a thread writes a key it checks that no other thread has written the key, and that no other threads read the key.

IPs/RSKIP144.md
ilanolkies: 👍

IPs/RSKIP144.md
```diff
@@ -9,48 +9,88 @@
...
+
+The REMASC transaction must be included as the last transaction of the sequential partition.
+
+An empty `partitionEnds` indicates that all transactions go in the _sequential partition_.
+
+Values in `partitionsEnd` must be greater than 0 and in ascending order. An empty `partitionEnds` indicates that all transactions go in the sequential partition. The maximum number of parallel partitions that the miner can specify is equal to the minimum number of cores required to run the RSK node, this maximum is called  `maxTransactionExecutionThreads`.
+
+## New block validation consensus
+
+The new consensus mechanism allows to reject blocks that, when executed in parallel, can result in different outputs.
```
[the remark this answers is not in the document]
ilanolkies: I want to say that there is a new way of marking blocks as invalid. This are new rules that didn't exist before, but do not mark as obsolete any old rule. Let me re-write this

IPs/RSKIP144.md
ilanolkies: I will leave this one as it is until we review and merge the current implementation. In any case, this is a valid _suggested_ implementation.

PR #334, opened 2022-06-30
Added tracking codes, favicons, and fixed dup. h1

max-barrera-2148:


index.html
```diff
@@ -30,6 +40,15 @@
...
   </script>
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-XT0RMBXMZM"></script>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
+    gtag('js', new Date());
+
+    gtag('config', 'G-XT0RMBXMZM');
+  </script>
```
ilanolkies, line 51: This appears twice. We are instancing it in head.html include too. I don't see where we are using `include head` so maybe it doesn't appear duplicated in the site. I would **make user we are using only one of inline head in index.html or including the head.html but not maintaining both files**

ilanolkies (changes requested):

ilanolkies:
Closing in favor of #342

PR #336, opened 2022-07-11
Simple Parallelizable Semaphore

SergioDemianLerner:


IPs/RSKIP336.md
```diff
@@ -0,0 +1,60 @@
...
+The only way to paralellize contracts using this semaphore contruction would be that the VM automatically detect that a variable is used as a sempahore, and proves that the sempahore behaves as required by static analysys. This is difficult for two reasons. First, the CPU cost of detection and/or proving may be high. Second, some contract functions may not be protected by the semaphore, meaning that some execution paths do not alter the semaphore, which may complicate proving. 
+Therefore this proposal aims to protect future contracts, rather than to protect already deployed ones.
+
+# **Specification**
+
+Starting from an activation block (TBD) a new precompiled contract `Semaphore` is created at address 0x01000011. when `Semaphore` is called, if the caller address is present more than once in the stack, the contract behaves as if the first instruction had been a REVERT, therefore the CALL returns 0. Otherwise, it executes no code and returns 1. The gas cost of the contract execution is set to 500, which is consumed independently of the call result.
```
ilanolkies, line 37:
```suggestion
Starting from an activation block (TBD) a new precompiled contract `Semaphore` is created at address 0x01000011. When `Semaphore` is called, if the caller address is present more than once in the stack, the contract behaves as if the first instruction had been a REVERT, therefore the CALL returns 0. Otherwise, it executes no code and returns 1. The gas cost of the contract execution is set to 500, which is consumed independently of the call result.
```

IPs/RSKIP336.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Rationale
+
+There are alternative designs to implement semaphores on the EVM such as 
+[EIP-1153: Transient storage opcodes](https://eips.ethereum.org/EIPS/eip-1153) or [`SSTORE_COUNT`](https://github.com/ethereum/EIPs/issues/119). The first does reduce the cost of using semaphores, but does not address the problem of parallelization. The second enables execution parallelization but it is more complex as it returns the number `SSTORE` opcodes executed, not the number of reentrant calls. Reentrancy must be deducted from this value. Also the count `SSTORE` count is not currently tracked by the RSK EVM. Also, it is discuraged to implement a new opcode that does not exists in Ethereum as it increases the chances of future incompatibilities. The presented solution is cleaner and simpler than all other alternatives.
+
```
ilanolkies, line 43: Why gas cost is 500?

ilanolkies (changes requested):
Excellent idea!

About robustness: what if you would like to use more than one semaphore? Maybe adding a semaphore id to the call and checking no pair (semaphore id, caller) is in the stack is an easy extension

IPs/RSKIP336.md
```diff
@@ -0,0 +1,60 @@
...
+# Rationale
+
+There are alternative designs to implement semaphores on the EVM such as 
+[EIP-1153: Transient storage opcodes](https://eips.ethereum.org/EIPS/eip-1153) or [`SSTORE_COUNT`](https://github.com/ethereum/EIPs/issues/119). The first does reduce the cost of using semaphores, but does not address the problem of parallelization. The second enables execution parallelization but it is more complex as it returns the number `SSTORE` opcodes executed, not the number of reentrant calls. Reentrancy must be deducted from this value. Also the count `SSTORE` count is not currently tracked by the RSK EVM. Also, it is discuraged to implement a new opcode that does not exists in Ethereum as it increases the chances of future incompatibilities. The presented solution is cleaner and simpler than all other alternatives.
```
julianlen, line 42: RSK EVM = RVM

IPs/RSKIP336.md
julianlen, line 37: Maybe I'm missing something, but it isn't clear to me how it works and how it prevents two contracts writing at the same storage key

IPs/RSKIP336.md
SergioDemianLerner: Do you think it's clearer if I write RVM ?

IPs/RSKIP336.md
SergioDemianLerner: This should be explained in the Rationale section. I will add it.

SergioDemianLerner: Added!

IPs/RSKIP336.md
SergioDemianLerner: If you mean "two parallel executing contracts writing to the same storage key" then this proposal does nothing to change that. That will be detected by the tracking of the storage key itself.
This RSKIP is not to prevent something, but to allow something. What is allowed is to make transaction dependency checks work only on storage keys, but not on the semaphore, which has a specific function that we know does not interfere with parallel processing.

ilanolkies (approved):
LGTM!

ilanolkies:
For the record: we are not considering the case of multiple semaphores since there is no clear use case

PR #349, opened 2022-08-31
Rskip144 audit review

julianlen:
TODO: A4. Waiting for answers


IPs/RSKIP144.md
```diff
@@ -21,7 +21,7 @@ Now, RSK nodes process transactions from blocks one by one, in the specified ord
...
+Miners must specify a valid execution plan that is included in the block header. We propose a greedy algorithm that produces a satisfactory plan. In our algorithm, miners are forced to serialize transaction execution to create blocks. at the same time they execute the transactions, they discover runtime key-access overlaps between transactions and build an execution plan. Miners can also use alternative methods tu produce optimized schedules. For a simpler overlap detection and to prevent DoS attacks, an additional part is added including all the transactions that could not be parallelized, that is executed after the execution of the parallel parts is completed. Once all transactions have been processed, the partition is created along with a schedule that determines which transactions belong in each part.
```
ilanolkies, line 24:
```suggestion
Miners must specify a valid execution plan that is included in the block header. We propose a greedy algorithm that produces a satisfactory plan. In our algorithm, miners are forced to serialize transaction execution to create blocks. At the same time they execute the transactions, they discover runtime key-access overlaps between transactions and build an execution plan. Miners can also use alternative methods tu produce optimized schedules. For a simpler overlap detection and to prevent DoS attacks, an additional part is added including all the transactions that could not be parallelized, that is executed after the execution of the parallel parts is completed. Once all transactions have been processed, the partition is created along with a schedule that determines which transactions belong in each part.
```

IPs/RSKIP144.md
```diff
@@ -61,7 +63,7 @@ A new constant `maxTransactionExecutionThreads` is specified. It determines the
 
 ## New block validation consensus
 
-It must be ensured that blocks that have transactions that are executed in parallel always produce the same output. Therefore, when a block is executed in parallel, nodes must verify the resulting world state is deterministic.
+It must be ensured that blocks that have transactions that are executed in parallel always produce the same output. Therefore, when a block is executed (either in parallel or sequentially) nodes must verify the resulting world state is deterministic as if it had been executed in parallel.
```
ilanolkies, line 66:
```suggestion
It must be ensured that blocks that have transactions that are executed in parallel always produce the same output. Therefore, when a block is executed (either in parallel or sequentially) nodes must verify the resulting world state is deterministic as if it had been executed sequentially.
```

IPs/RSKIP144.md
```diff
@@ -73,8 +75,8 @@ Any pair of transactions that are _connected_ cannot be in different parallel su
 
 > Recursive deletes must be correctly and efficiently handled. If a transaction deletes a contract using `SELFDESTRUCT` and another transaction is reading or writing a key of that contract, those two transactions must be considered _connected_.
 
-This cases should not be considered as writing a key, since they doesn't affect the output of the parallel execution:
-- Adding 0 balance
+It is worth it mention that sending 0 balance to a non-existent account creates that account. Hence, a key is considered written.
```
ilanolkies, line 78: Maybe as a comment?

```suggestion
> Sending 0 balance to a non-existent account creates that account. Hence, a key is considered written.
```

ilanolkies (changes requested):
👌🏻

ilanolkies (approved):
Ok 👍

PR #374, opened 2023-02-15
Restore PowPeg migration age to its original value

marcos-iov:


IPs/RSKIP374.md
```diff
@@ -0,0 +1,50 @@
...
+|**Complexity** |1 |
+|**Status**     |Draft |
+
+## Abstract
+
+This RSKIP proposes a change to the number of block confirmations required for a PowPeg migration period.
+
+## Motivation
+
+As described in RSKIP357 [[1]](#references), as part of the efforts done to solve the peg-outs service incident[[2](#references)] it was necessary to increase the PowPeg migration period from 10,585 to 172,800 blocks to give enough time for a mining pool to process the non-standard transactions before the migration period finishes.
```
nathanieliov, line 29: This link references([[1]](#references), [[2](#references)]) don't take to any different place rather than the same page, because after the redirect to the link, the difference is almost imperceptible, seems like you stay on the same place.

What if we add the real reference? So it would look like this:

[1](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP357.md)

nathanieliov (approved):
LGTM

IPs/RSKIP374.md
```diff
@@ -0,0 +1,50 @@
...
+
+Once the issue with the peg-outs has been solved and the service has been resumed, this RSKIP proposes to restore the PowPeg migration period back to its orginal value of 10,585 blocks.
```
jeremy-then, line 31: original*

jeremy-then (approved):
LGTM.

IPs/RSKIP374.md
marcos-iov: Perhaps I could just remove the link then. And leave the reference as it is [1]. What do you think?

kelvinator07 (approved):

josedahlquist (approved):

PR #387, opened 2023-05-19
Support for Bridging Ordinals in RSK Bridge 

SergioDemianLerner:


IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+To obtain the ordFedAddress, the following bridge method must be called:
+
+`String getFederationOrdinalsAddress()`
+
+The address is constructed differently from the BTC peg-in address. The construction is described in the section “Main Ordinals Address Derivation”
+
+To register an ordinal, the following method can be called:
+
+`void registerBtcTransactionWithOrdinals(bytes rskTx, bytes btcTxSerialized, int height, bytes pmtSerialized)`
```
josedahlquist, line 74: What's the purpose of the `rskTx` argument?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+The method does nothing if the ordinal does not exist.
+
+## Ordinal Peg-outs
+
+To send an ordinal, the following method must be called:
+
+`void releaseOrdinal(uint256 txId, uint256 index, bytes btcAddress)`
+
+This method will queue the release of an ordinal, sending it to the btcAddress specified. The caller must be the owner of the ordinal. Every 1 hour the ordinals that are queued to be released will be processed. The bridge may attempt to create a single transaction containing all ordinal UTXOs, and all ordinal outputs, plus a single additional input used for fee payments, and a single additional output used for change.
```
josedahlquist, line 117: Who pays for the peg-out fees in this case? Are we considering that the parking fees should cover for the peg-out costs?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Security 
+
+We believe that while some ordinals and inscriptions could be highly valued, due to the low cost of creating spammy inscriptions, in the long term most ordinals will be low valued. The cost of an ordinal may become much lower than the BTC amount of an average bitcoin peg-in, and even lower than the cost of a UTXO consolidation. This has two consequences: first, ordinals should pay a rent to avoid spam, and second, ordinals can be carefully isolated with their own security model. 
+While the double-spend of an ordinal during peg-in can cause distress to its owner, it does not affect any other user of Roostock, not any other ONFT. There can’t be a run to the peg if an ordinal is lost, while a mismatch of BTC collateral in Bitcoin and rBTC  issued in Roostock can be the cause of such an event. 
+
+Since ordinals can be tracked independently in separate UTXOs, the double-spend of an ordinal does not affect the peg funds, nor the other ordinals. Therefore, we can lower the number of confirmations required for an ordinal peg-in. We propose that ordinals require 6 Bitcoin confirmations for peg-in (about 1 hour), and 120 confirmations (about 1 hour) for peg-out.
```
josedahlquist, line 229: This will imply a change in POWHSM as well. It cannot discern between regular peg-outs and ordinal peg-outs.
Should it?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+The Rootstock bridge is designed to transfer fungible tokens, but not non-fungible elements. To transfer non-fungible elements, the bridge must segregate them, and avoid mixing them with other elements. The bridge is not prepared to do this now, and sending an UTXO containing an ordinal to the Bridge will cause the ordinal to be commingled with other UTXOs, and its ownership would be inadvertently transferred to a user performing a peg-out at a later time.
+
+# **Specification**
+
+This is a proposal for making Bitcoin ordinals and inscriptions available in Rootstock using the current Powpeg bridge. Once ordinals or inscriptions have representations in Rootstock,  they can be traded as ONFTs (Ordinal NFTs). At any moment, the owner of the ONFTs can redeem them in Bitcoin and transfer them to any Bitcoin address.
+
+Internally, the bridge does not know anything about NFTs or EIP-712. The bridge will only provide the service of the temporal custody of isolated UTXOs, what we’ll call parking. The bridge will provide one additional service: changing the owner of a parked UTXO.
```
kelvinator07, line 40:
```suggestion
Internally, the bridge does not know anything about NFTs or EIP-721. The bridge will only provide the service of the temporal custody of isolated UTXOs, what we’ll call parking. The bridge will provide one additional service: changing the owner of a parked UTXO. 
```

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+# **Specification**
+
+This is a proposal for making Bitcoin ordinals and inscriptions available in Rootstock using the current Powpeg bridge. Once ordinals or inscriptions have representations in Rootstock,  they can be traded as ONFTs (Ordinal NFTs). At any moment, the owner of the ONFTs can redeem them in Bitcoin and transfer them to any Bitcoin address.
+
+Internally, the bridge does not know anything about NFTs or EIP-712. The bridge will only provide the service of the temporal custody of isolated UTXOs, what we’ll call parking. The bridge will provide one additional service: changing the owner of a parked UTXO. 
+
+Using external user-provided smart-contracts, those ordinals can be turned into EIP-712-compatible NFTs. To enable this, the community needs to define a standard URI for the lookup of ordinals in the Bitcoin blockchain, so that NFTs marketplaces can retrieve the associated inscription. That standard is out of scope of this document.
```
kelvinator07, line 42:
```suggestion
Using external user-provided smart-contracts, those ordinals can be turned into EIP-721-compatible NFTs. To enable this, the community needs to define a standard URI for the lookup of ordinals in the Bitcoin blockchain, so that NFTs marketplaces can retrieve the associated inscription. That standard is out of scope of this document.
```

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+To avoid attackers spamming the Bridge with ordinals, there are upfront and recurrent costs for parking ordinals within RSK. Ordinals whose annual fee has been unpaid will be assigned to the first user who pays its parking fees. In the future, a blinded auction process could be added.  
+ 
+The Bridge will provide these new methods:
+
+* **getFederationOrdinalsAddress**(): get the current address to transfer ordinals to Rootstock.
+* **registerBtcTransactionWithOrdinals**(): registers all the ordinals in a transaction. The ordinals are identified by being sent to the ordinals federation address.
+* **registerBtcTransactionWithOrdinalsDMFA**(): registers all ordinals in a transaction. Ordinals are identified by being sent to unique federation addresses mapped to a destination owner.
+* **getOrdinalOwner**(): Returns the RSK owner of an ordinal. 
+* **transferOrdinal**(): sets a new RSK owner for an ordinal. Only the owner can transfer ownership of  an ordinal.
+* **releaseOrdinal**(): transfers the ordinal back to a Bitcoin user of the Bitcoin blockchain. Only the owner can release an ordinal.
```
marcos-iov, line 53: Should we use the term `pegout` instead of release to keep it consistent with the 2wp?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Ordinal Peg-outs
+
+To send an ordinal, the following method must be called:
+
+`void releaseOrdinal(uint256 txId, uint256 index, bytes btcAddress)`
+
+This method will queue the release of an ordinal, sending it to the btcAddress specified. The caller must be the owner of the ordinal. Every 1 hour the ordinals that are queued to be released will be processed. The bridge may attempt to create a single transaction containing all ordinal UTXOs, and all ordinal outputs, plus a single additional input used for fee payments, and a single additional output used for change.
+
+Ordinals can be sent to any Bitcoin address.
```
marcos-iov, line 119: The Bridge currently does not support creating transactions to segwit addresses. Should that be a pre-requisite or should we add a limitation to the type of recipient addresses?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+This method will register the UTXOs as independent entities. Each UTXO registered will be assigned a serial number (or index) starting from 0 on each transaction.  This serial number matches the UTXO output index. If there are outputs that are not transfers to the federation ordinal address, then those serial numbers will be skipped.
+
+The UTXOs will be owned in RSK by the same private key that controls the first input of the transaction or owned by a different RSK address, if there is an `OP_RETURN` output containing it, as specified by RSKIP-170.
+
+
+## Peg-In using Destination-Mapped Federation Addresses
+
+It’s also possible to create a Bitcoin address that automatically assigns the ownership of the ordinal in Rootstock to a specific RSK address or contract. This is called destination-mapped federation addresses (DMFA). This can be used to automatically mint an NFT when the ordinal is transferred to Rootstock. 
+
+This requires creating a special unique single-use reception address for the ordinal. The address is created with the following method:
```
marcos-iov, line 85: The method below should return a Bitcoin address that when ordinals are sent to it they should be assigned to the `rskOwner`, is that how it works? So after getting this address the user needs to call `registerBtcTransactionWithOrdinals`, is that right?

What's the difference between doing it this way and using RSKIP 170 specs?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+Ordinals can be sent to any Bitcoin address.
+
+## Paying Parking Fees
+Each ordinal parked stores the date when the last annual parking fee was paid. Initially this date corresponds to the date of parking. The parking expires one year later. If the parking expires and the fee has not been paid, anyone will be able to claim the ordinal by paying its parking fee. If a bridge migration event occurs, all parked ordinals that have their fee unpaid will not be migrated, and they will be lost forever.
+
+To know how much must be paid to park an ordinal, the Bridge provides the following method: 
+
+`uint getOrdinalsAnnualParkingFee()`
+
+This method returns the fee that ordinals must pay annually for being parked. The amount is given in rBTC weis.
```
marcos-iov, line 128: Is this a fixed value?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+To send an ordinal, the following method must be called:
+
+`void releaseOrdinal(uint256 txId, uint256 index, bytes btcAddress)`
+
+This method will queue the release of an ordinal, sending it to the btcAddress specified. The caller must be the owner of the ordinal. Every 1 hour the ordinals that are queued to be released will be processed. The bridge may attempt to create a single transaction containing all ordinal UTXOs, and all ordinal outputs, plus a single additional input used for fee payments, and a single additional output used for change.
+
+Ordinals can be sent to any Bitcoin address.
+
+## Paying Parking Fees
```
marcos-iov, line 121: When the ordinal is first registered it will initially show as unpaid, right? So the user should call `payParkingExtensionFee` right after calling the register method

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+To pay the parking fee of an ordinal, the Bridge provides the following method:
+
+`void payParkingExtensionFee(uint256 tId, int index)`
+
+This method extends the parking time of the ordinal to a year starting from the current date.
+ 
+To know exactly how much you need to pay to extend the parking time, the Bridge provides the following method:
+ 
+`uint256 getOrdinalParkingExtensionFee(uint256 tId, int index)`
```
marcos-iov, line 138: Is this value different from `getOrdinalsAnnualParkingFee`? It's still the fee for parking the ordinal for a year, isn't it?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+This method returns how much the user needs to pay to extend the parking period for one year from the current date. The value is expressed in weis.
+
+To know when an ordinal parking period expires, the Bridge provides this method:
+
+`void getOrdinalParkingExpirationDate(uint256 tId, int index)`
+
+
+When paying the parking extension fee, the user should transfer the amount returned by getOrdinalParkingExtensionFee(). Any excess amount will be reimbursed to the sender by direct balance update. Note that if the user pays the parking fee several times in the same block, only the first payment will be processed. The remaining will be fully reimbursed.
```
marcos-iov, line 147: Could a user pay the parking fee for several years in advance?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Single-use Ordinals Address Derivation
+The Powpeg ordinals address derivation starts with the creation of a custom redeem script that pushes the verOwner field. The verOwner field is defines as:
+
+* **version**: specifies the derivation version (1 byte) 
+* **owner**: ordinal owner address (a 20 byte RSK address)
+
+Therefore verOwner consist of a 21-byte binary chunk. The verOwner field is pushed and immediately dropped from the stack using the `OP_DROP` opcode, so the pushed data does not alter the processing of the redeem script.
+
+The redeem script for Powpeg-derived addresses is a modification of the standard Powpeg redeem script. We show how to build such scripts, but we omit the emergency multisig script path for clarity. The whole script must be considered when computing the address. The new custom redeem script has the following format:
```
marcos-iov, line 194: Just to be clear, the ordinal address redeem script will include the emergency multisig as well, right?

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+|**Author**     |SDL |
+|**Purpose**    |Usa |
+|**Layer**      |Core |
+|**Complexity** |2 |
+|**Status**     |Draft |
+|**discussions-to**     ||
+
+# **Abstract**
+
+This is a proposal for making Bitcoin ordinals and inscriptions available in Rootstock using the current Powpeg bridge. Once ordinals or inscriptions have representations in Rootstock,  they can be traded. Later they can be transfered back to any Bitcoin address.
```
nathanieliov, line 27:
```suggestion
This is a proposal for making Bitcoin ordinals and inscriptions available in Rootstock using the current Powpeg bridge. Once ordinals or inscriptions have representations in Rootstock,  they can be traded. Later they can be transferred back to any Bitcoin address.
```

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+
+# **Motivation**
+
+Inscriptions are images, videos, recordings or text that is stored in the Bitcoin blockchain, can be owned by a user and the ownership can be transferred to another user using standard Bitcoin transactions. We can think that every satoshi ever minted in Bitcoin is a separate “virtual cheque” and has an unique serial number. Then those serial numbers are called Ordinals. Inscriptions are linked to specific ordinals, and move with those ordinals with every transaction that involves those virtual cheques. There has been a rising interest on inscriptions since their introduction.
```
nathanieliov, line 32:
```suggestion
Inscriptions are images, videos, recordings, or text that are stored in the Bitcoin blockchain, can be owned by a user and the ownership can be transferred to another user using standard Bitcoin transactions. We can think that every satoshi ever minted in Bitcoin is a separate “virtual cheque” and has a unique serial number. Then those serial numbers are called Ordinals. Inscriptions are linked to specific ordinals, and move with those ordinals with every transaction that involves those virtual cheques. There has been a rising interest on inscriptions since their introduction.
```

IPs/RSKIPxxx.md
SergioDemianLerner: The current bridge method for peg-out is called releaseBtc(), so I thought "release" was more consistent.

IPs/RSKIPxxx.md
SergioDemianLerner: typo. copy-pasted from bridge code by mistake.

IPs/RSKIPxxx.md
SergioDemianLerner: Yes. The workflow is similar to RSKIP170.
The explanation is incorrect. This method does not return an address. The method actually registers the ordinal.

SergioDemianLerner: I will improve the text.

IPs/RSKIPxxx.md
SergioDemianLerner: Yes, for simplification, the parking fee should cover the peg-out costs. Do you think it should be paid on peg-out ?

IPs/RSKIPxxx.md
SergioDemianLerner: We can add the limitation. But I think that sending to any Bitcoin address would not be very complex to implement. It's just parsing the address and converting it to the correct scriptpub.

IPs/RSKIPxxx.md
SergioDemianLerner: Let me clarify this: the powpeg federation is NOT involved in any way in registering the ordinals. 

It's up to the user to perform the corresponding RSK transaction to register them. The first year of parking is paid when registering. This is missing in the text and it will be added.

IPs/RSKIPxxx.md
SergioDemianLerner: A multiple of  fee x kilobyte value already configured.
It will change if the feeXkilobyte is changed

IPs/RSKIPxxx.md
SergioDemianLerner: This is the extension cost. If currently the expiration is in 6 months, then the value returned will be half of the annual cost, because 6 months have already been paid and only 6 months remain to be paid. Remember that the rent cannot be paid in advance for more than a year.

IPs/RSKIPxxx.md
SergioDemianLerner: No. Currently the maximum that can be paid is one year.

IPs/RSKIPxxx.md
SergioDemianLerner: Do you have an idea of how easy/hard is to allow this? Can the HSM distinguish between an ordinal peg-out and a btc peg-out ? It could use the output value to distinguish, yet there is the danger that it accepts a 120-conf peg-out that drains all the funds in miner fees, so it must also use the amount of the fees to distinguish, which makes the process much more complicated, because the HSM doesn't know the amounts of the inputs. Input amounts are only available explicitly in segwit transactions.  

A better and more secure option is that the Powpeg uses a different set of public keys to manage ordinals, so one set of keys manage 120-conf peg.outs, and another set, 4000-conf peg.outs.

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+`void registerBtcTransactionWithOrdinals(bytes rskTx, bytes btcTxSerialized, int height, bytes pmtSerialized)`
+
+This method will register the UTXOs as independent entities. Each UTXO registered will be assigned a serial number (or index) starting from 0 on each transaction.  This serial number matches the UTXO output index. If there are outputs that are not transfers to the federation ordinal address, then those serial numbers will be skipped.
+
+The UTXOs will be owned in RSK by the same private key that controls the first input of the transaction or owned by a different RSK address, if there is an `OP_RETURN` output containing it, as specified by RSKIP-170.
+
+
+## Peg-In using Destination-Mapped Federation Addresses
+
+It’s also possible to create a Bitcoin address that automatically assigns the ownership of the ordinal in Rootstock to a specific RSK address or contract. This is called destination-mapped federation addresses (DMFA). This can be used to automatically mint an NFT when the ordinal is transferred to Rootstock.
```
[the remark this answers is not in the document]
SergioDemianLerner: The workflow would be similar to the current flyover. A contract must call the bridge.

IPs/RSKIPxxx.md
```diff
@@ -0,0 +1,262 @@
...
+
+The address is constructed differently from the BTC peg-in address. The construction is described in the section “Main Ordinals Address Derivation”
+
+To register an ordinal, the following method can be called:
+
+`void registerBtcTransactionWithOrdinals(bytes rskTx, bytes btcTxSerialized, int height, bytes pmtSerialized)`
+
+This method will register the UTXOs as independent entities. Each UTXO registered will be assigned a serial number (or index) starting from 0 on each transaction.  This serial number matches the UTXO output index. If there are outputs that are not transfers to the federation ordinal address, then those serial numbers will be skipped.
+
+The UTXOs will be owned in RSK by the same private key that controls the first input of the transaction or owned by a different RSK address, if there is an `OP_RETURN` output containing it, as specified by RSKIP-170.
```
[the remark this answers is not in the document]
SergioDemianLerner: To make it simple, there are no rejections. Either the user includes a valid address in an OP_RETURN output or a valid P2PKH first input.
We can enable the ordinal to be claimed by anyone after 6 months. Do you think this would cover this case ?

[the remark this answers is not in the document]
SergioDemianLerner: Regarding spamming attack: there can't be any, because the registration of the ordinal is made by the user with an RSK transaction.

IPs/RSKIPxxx.md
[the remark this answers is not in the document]
SergioDemianLerner: No RSK call is triggered by the BTC transaction

IPs/RSKIPxxx.md
SergioDemianLerner: clarified

IPs/RSKIP387.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Ordinal Peg-outs
+
+To send an ordinal, the following method must be called:
+
+`function releaseOrdinal(uint256 txId, uint256 index, bytes btcScriptPubkey)`
+
+This method will queue the release of an ordinal, sending it to the Bitcoin address associated with the btcScriptPubkey specified. The caller must be the owner of the ordinal. Every 1 hour the ordinals that are queued to be released will be processed. The bridge may attempt to create a single transaction containing all ordinal UTXOs, and all ordinal outputs, plus a single additional input used for fee payments, and a single additional output used for change.
+
+Ordinals can be sent to any Bitcoin address by directly specifying the btcScriptPubkey, and the called must make sure that the btcScriptPubkey given is the correct expansion of a standard, Bech32 or any other future Bitcoin address. The Bridge will not perform any validation on the btcScriptPubkey. If it is invalid, the ordinal will be lost.
```
SergioDemianLerner, line 122: I see a problem with this and peg-out batching: what if the btcScriptPubkey makes the transaction non-standard ? We would need to perform some minimal validation on the scriptPub to avoid it.

SergioDemianLerner: @josedahlquist Do you agree?

IPs/RSKIP387.md
```diff
@@ -0,0 +1,303 @@
...
+* **payParkingExtensionFee**(): Pays the parking fee of an ordinal to be extended one year. 
+* **getOrdinalParkingExtensionFee**(): Returns how much the user needs to pay to extend the parking period for one year from the current date.
+* **getOrdinalParkingExpirationDate**(): Returns the date the current parking period ends.
+
+Starting from an activation block (TBD) these methods are activated.
+ 
+
+## Peg-in of an ordinal
+
+The bridge has a special address to send ordinals or inscriptions: the ordFedAddress. Sending BTCs in ordinals UTXOs with amounts over 0.01 BTC is forbidden. In the case it happens, the Bridge will migrate the funds to the normal Powpeg address, and consider tthe funds a donation to the Rootstock sidechain. The control of the associated ordinal will be lost.
```
ilanolkies, line 66:
```suggestion
The bridge has a special address to send ordinals or inscriptions: the ordFedAddress. Sending BTCs in ordinals UTXOs with amounts over 0.01 BTC is forbidden. In the case it happens, the Bridge will migrate the funds to the normal Powpeg address, and consider the funds a donation to the Rootstock sidechain. The control of the associated ordinal will be lost.
```

smishraIOV (approved):

PR #417, opened 2024-02-08
Improve empty Bridge response when called from a smart contract

marcos-iov:


IPs/RSKIP417.md
```diff
@@ -0,0 +1,46 @@
...
+purpose: Usa
+layer: Core 
+complexity: 1
+status: Draft
+description: 
+---
+
+|RSKIP          |417           |
+| :------------ |:-------------|
+|**Title**      |Improve empty Bridge response when called from a smart contract |
```
aeidelman, line 15: Suggesting another name: Avoid transactions to be reverted when Bridge method calls from smart contracts return an empty response. Is it too long?

aeidelman (approved):
LGTM

PR #427, opened 2024-04-17
Express amount value as weis in release related events

marcos-iov:


jeremy-then (approved):
LGTM.

IPs/RSKIP427.md
```diff
@@ -0,0 +1,45 @@
...
+
+This RSKIP proposes a change in the unit used in `amount` field of `release_request_received` and `release_request_rejected` Bridge events. The proposal is to express the amount in **weis** instead of **satoshis**.
+
+## Motivation
+
+In the original specification [1] it was specified that `release_request_received` and `release_request_rejected` Bridge events should express the _amount_ field in weis.
+
+## Specification
+
+When `release_request_received` or `release_request_rejected` Bridge events are emitted, the _amount_ field should include the value in weis that the user sent to the Bridge.
```
nathanieliov, line 33: Should we mention that the event signature will remain as it is? What do you think?

nathanieliov (commented):
LGTM, just a comment/suggestion.

IPs/RSKIP427.md
julia-zack, line 33: I would say `the _amount_ field should be the value that the user sent to the Bridge **expressed in weis**` or something like that.
For me, the way it's currently said sounds like there is no amount being included in the event.

IPs/RSKIP427.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Backward Compatibility
+
+This change is a hard fork and therefore all full nodes must be updated.
```
julia-zack, line 37: We may include `Since the same amount is different if expressed in weis rather than satoshis, this change is a hard fork...`
Just a thought

IPs/RSKIP427.md
marcos-iov: Added, let me know what you think

IPs/RSKIP427.md
```diff
@@ -0,0 +1,45 @@
...
+|**Complexity** |1 |
+|**Status**     |Draft |
+
+## Abstract
+
+This RSKIP proposes a change in the unit used in `amount` field of `release_request_received` and `release_request_rejected` Bridge events. The proposal is to express the amount in **weis** instead of **satoshis**.
+
+## Motivation
+
+In the original specification [1] it was specified that `release_request_received` and `release_request_rejected` Bridge events should express the _amount_ field in weis.
```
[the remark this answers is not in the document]
marcos-iov: Changed wording a bit

IPs/RSKIP427.md
marcos-iov: Changed wording, let me know what you think

IPs/RSKIP427.md
marcos-iov: Mmm not sure, I think that should be already understood by the specification

nathanieliov (approved):
LGTM

apancorb (approved):
LGTM, just a question. What would be some potential issues of having the value in satoshis rather than weis? Is it a small mismatch in the bridged amount?

IPs/RSKIP427.md
```diff
@@ -0,0 +1,45 @@
+---
+rskip: 427
+title: Express amount value as weis in peg-out related events
```
aeidelman, line 3: Minor change suggested: "Express the amount value in wei for peg-out related events."

aeidelman (changes requested):
Suggested a small change in the RSKIP title. Other than that, looks good

IPs/RSKIP427.md
marcos-iov: Updated, thanks!

wilmerrootstock (approved):
LGTM!

aeidelman (approved):

PR #432, opened 2024-05-24
Add RSKIP for EthSwap proposal

rmoreliovlabs:


IPs/RSKIP432.md
```diff
@@ -0,0 +1,69 @@
+---
+rskip: 432
+title: RbtcSwap
+description:
+status: Draft
+purpose: Usa
+author: (@rmoreliovlabs)
```
aeidelman, line 7: Should be "RM" and add you to the README file as well

IPs/RSKIP432.md
```diff
@@ -0,0 +1,69 @@
...
+created: 2024-05
+---
+# RbtcSwap
+
+
+|RSKIP          | 432 |
+| :------------ |:-------------|
+|**Title**      |RbtcSwap|
+|**Created**    |MAY-2024 |
+|**Author**     |@rmoreliovlabs |
```
aeidelman, line 19: Same as before

IPs/RSKIP432.md
```diff
@@ -0,0 +1,69 @@
+---
+rskip: 432
+title: RbtcSwap
```
aeidelman, line 3: Can we have a little bit more descriptive name? Something like "Add new RBTCSwap contract"

aeidelman (changes requested):

IPs/RSKIP432.md
rmoreliovlabs: How about the name `RbtcSwap mechanism`? This feature is not a new contract, but an implementation of a mechanisim that uses an existing contact

IPs/RSKIP432.md
aeidelman: I see you included RM in the README but we also use initials here

IPs/RSKIP432.md
aeidelman: But it's a hardcoded bytecode contract that you invoke how ?

rmoreliovlabs: Sorry if it got confusing, what I meant is that this RSKIP is not only to add the contract but also the mechanism that allows the claim to be paid after its execution.

PR #473, opened 2025-01-24
Add latest changes to RSKIP419

julia-zack:


IPs/RSKIP419.md
```diff
@@ -208,9 +211,9 @@ The duration of this phase must account for:
 - Sending back the UTXO to the current Powpeg (Proposed pegnatories signing)
 
 The recommendation is that this phase takes approximately the blocks a peg-out confirmation has plus the bitcoin blocks needed to register the change, twice. And a couple of thousands more blocks to be safe.
-- Mainnet: 16_000
-- Testnet: 80 - For testnet it is recommended to wait a bit longer as the Bitcoin miners don't tend to follow the 10' block creation time.
-- Regtest: 15
+- Mainnet: 20_000
+- Testnet: 2_000 - For testnet it is recommended to wait a bit longer as the Bitcoin miners don't tend to follow the 10' block creation time. _Disclaimer: this implies increasing the federation activation age value also, since the validation period has to be lower than the activation one._
```
marcos-iov, line 215: Let's be clear what the new value is

IPs/RSKIP419.md
```diff
@@ -247,10 +250,15 @@ getStateForSvpClient() returns bytes
 commitFederationFailed(bytes proposedFederationRedeemScript, uint blockNumber)
 ```
 
+### Other changes
+To follow EVM common practices, the `getProposedFederationCreationTime` Bridge's method returns the value from the epoch seconds.
```
marcos-iov, line 254:
```suggestion
To follow EVM common practices, `getProposedFederationCreationTime` Bridge method returns the value from the epoch seconds.
```
I think that's the correct way. `Bridge method` instead of `Bridge's method`

IPs/RSKIP419.md
```diff
@@ -247,10 +250,15 @@ getStateForSvpClient() returns bytes
...
+To be consistent with this, the creation time being returned from `getFederationCreationTime` and `getRetiringFederationCreationTime` Bridge's methods is modified to be this way.
```
marcos-iov, line 255:
```suggestion
To be consistent with this, the creation time being returned from `getFederationCreationTime` and `getRetiringFederationCreationTime` Bridge methods is modified to also return the value from the epoch seconds instead of milliseconds.
```
Feel free to change it, jus think we should be explicit on what the change is

IPs/RSKIP419.md
julia-zack: updated

IPs/RSKIP419.md
julia-zack: yep, updated!

IPs/RSKIP419.md
julia-zack: updated

PR #502, opened 2025-03-19
Union Bridge - PowPeg integration

marcos-iov:


IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+
+**Storage entry name:** weisTransferredToUnionBridge
+
+Tracks the total amount of RBTC (expressed in weis) transferred to the Union Bridge.
+
+### 7. Receiving funds back from the Union Bridge
+
+The Union Bridge contract will be capable of sending funds back to the PowPeg. When this happens, the tracking entry for the transferred amount will need to be updated.
+
+No new method will be added. It will require modifying the existing `releaseBtc` method to include special logic for transactions received from the Union Bridge contract address.
```
SergioDemianLerner, line 129: I think that for symmetry there should be a specific releaseUnionRBTC() method.

IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+### 5. Method for Transferring RBTC to the Union Bridge
+
+A new method will be added to the Bridge contract to allow the transfer of RBTC to the Union Bridge. The method will have the following restrictions:
+
+- Only callable by the Union Bridge contract.
+- The total requested amount (current request + previously requested amounts) must not exceed the current locking cap.
+
+**Method Signature:** 
+```
+function requestRBTC(uint256 amountInWeis) public returns
```
SergioDemianLerner, line 108: The method should be called requestUnionRBTC() to avoid confusion and match a specific method used to release of btc

IPs/RSKIP502.md
SergioDemianLerner: Also an event should be triggered.

IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+An event will be emitted each time this method is invoked, to provide transparency about the transfer.
+
+```
+event rbtc_requested(address indexed requester, uint256 amountInWeis);
```
SergioDemianLerner, line 114: event name should be clearer: union_rbtc_requested

IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+
+A new storage entry will be used to track the total amount of RBTC that has been transferred to the Union Bridge. This is essential to ensure that the locking cap is not exceeded.
+
+**Storage entry name:** weisTransferredToUnionBridge
+
+Tracks the total amount of RBTC (expressed in weis) transferred to the Union Bridge.
+
+### 7. Receiving funds back from the Union Bridge
+
+The Union Bridge contract will be capable of sending funds back to the PowPeg. When this happens, the tracking entry for the transferred amount will need to be updated.
```
SergioDemianLerner, line 127: add: the bridge will not allow weisTransferredToUnionBridge to go negative, which could only happen if the union bridge is malfunctioning. If this happens, the method will fail with an error code and  the bridge will automatically enter the pause mode.
The best path forward in this case we can discuss.

IPs/RSKIP502.md
SergioDemianLerner: having a specific method allows to support specific error codes also.

IPs/RSKIP502.md
pmprete: Also we could add a destination address so we can send tokens directly. That way we will save gas, avoiding  having to send the tokens to the union bridge and then to the user.
 Should end up looking something like this
 
```suggestion
function requestUnionRBTC(address to, uint256 amountInWeis) public returns
```

IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+```
+function requestRBTC(uint256 amountInWeis) public returns
+```
+
+An event will be emitted each time this method is invoked, to provide transparency about the transfer.
+
+```
+event rbtc_requested(address indexed requester, uint256 amountInWeis);
+```
+
```
pmprete, line 116: We should mention the response codes, something like

```md
**Response codes:**
- 0: Success – The Bridge successfully transfered RBTC.
- -1: Unauthorized caller – The caller should be the union bridge contract address.
- -2: Locking cap exceeded – The caller should be the union bridge contract address.
- -10: Generic error – A non-specific error occurred while processing the request.
```

IPs/RSKIP502.md
josedahlquist: I like Peter's idea about transferring the rbtc directly but we must not forget the bridge doesn't do a transfer per but instead it changes the balance, therefore if the recipient is a contract it won't execute any code

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+## Motivation
+
+The Union Bridge is a new project currently under development that will interact with the Rootstock network. Integrating the Union Bridge with the existing PowPeg is essential to facilitate the transfer of RBTC. A key concern is ensuring that the integration remains secure, and that risks are minimized through a locking cap that controls the amount of RBTC transferred. The PowPeg's current functionality must be extended with minimal disruption, while also accounting for the need to manage the evolving nature of the Union Bridge.
+
+Additionally, the Union Bridge is still in development, and its contract address is likely to change as it evolves. To accommodate this, the proposed solution includes a flexible mechanism for updating the Union Bridge contract address during its testing and development phases, ensuring that these changes can occur securely without risking mainnet operations.
```
julia-zack, line 31: Will we consider that the contract address could change in a future, after the "development phase"?

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+
+This proposal aims to ensure that the PowPeg remains secure while enabling a smooth transition and integration with the Union Bridge as it matures, all while minimizing risks and maintaining full compatibility with the current system.
+
+## Specification
+
+### 1. New storage entry for Union Bridge contract address
+
+A new storage entry is necessary to store the address of the Union Bridge contract. This will allow the Bridge to identify when transactions are coming from the Union Bridge, and provide a mechanism for managing these interactions securely.
+
+Storage entry name: `unionBridgeContractAddress`
```
julia-zack, line 41:
```suggestion
Storage entry name: `unionBridgeAddress`
```
maybe theres no need of the _contract_ word

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+
+Maximum size: 32 bytes
```
apancorb, line 43: Why is it 32 bytes? Should this be 20 bytes?

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+This method will be only enabled for testnet and regtest environments.
+It will be disabled on mainnet to prevent unauthorized updates.
+
+### 3. Authorizer for changing the Union Bridge contract address
+
+An authorizer key will be created to allow the update of the Union Bridge contract address. This key will be similar to the whitelist authorizer [[2]](#references), with a single key being sufficient for this purpose. No voting or multisig will be needed since this will only be relevant on testnet.
+
+### 4. Union Bridge locking cap
+
+A locking cap will be implemented to control the maximum amount of RBTC that can be transferred to the Union Bridge. This cap helps mitigate risks by preventing excessive transfers beyond the system's control limits.
```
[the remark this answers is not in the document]
apancorb: I agree, it could be helpful.

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+Tracks the total amount of RBTC (expressed in weis) transferred to the Union Bridge.
+
+### 7. Receiving funds back from the Union Bridge
+
+The Union Bridge contract will have the capability to send funds back to the PowPeg. When this happens, the tracking entry for the amount transferred will need to be updated to reflect the returned RBTC. The Bridge will not allow **weisTransferredToUnionBridge** to go negative, which could only happen if the Union Bridge is malfunctioning. If this happens, the method will fail with an error code and the Union-PowPeg integration will automatically enter pause mode.
+
+A new method will be added to the Bridge contract to allow the transfer of RBTC from the Union Bridge back to the PowPeg. The method will have the following restrictions:
+
+- It will only be callable by the Union Bridge contract address.
+- The amount sent back cannot exceed the total amount of RBTC previously transferred to the Union Bridge, as tracked by the **weisTransferredToUnionBridge** storage entry. In other words, the Union Bridge cannot return more RBTC than it has previously received.
```
apancorb, line 146: Since we are dealing with RBTC should the value be measured in sats rather than weis?

IPs/RSKIP502.md
nathanieliov: Agree. Good catch.

IPs/RSKIP502.md
marcos-iov: Not at this stage, the Union Bridge contract will probably be a proxy

IPs/RSKIP502.md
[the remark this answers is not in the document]
marcos-iov: TBD

IPs/RSKIP502.md
marcos-iov: Good catch, thanks

IPs/RSKIP502.md
```diff
@@ -0,0 +1,252 @@
...
+Maximum size: 32 bytes
+
+### 2. Logic for updating the Union Bridge contract address
+
+The Union Bridge contract address will be updated periodically during its development and testing phase. A new method will be added to the Bridge contract to allow updating this address. This method will only be available on testnet and regtest environments and will not be enabled on mainnet.
+
+**Method signature:**
+
+```
+function setUnionBridgeContractAddressForTestnet(address unionBridgeContractAddress) public returns (int256);
```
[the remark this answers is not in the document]
marcos-iov: Good idea, added

IPs/RSKIP502.md
```diff
@@ -0,0 +1,175 @@
...
+
+To allow for emergency halts or suspension of interactions between the PowPeg and the Union Bridge, a pause mechanism will be implemented. When the system is paused, all transfers between the PowPeg and the Union Bridge will be suspended until the pause is lifted.
+
+The pause mechanism will allow authorized signers to pause the system temporarily. Once the system is paused, no transfers of RBTC between the PowPeg and Union Bridge can occur. The system can later be unpaused, resuming the normal transfer of RBTC.
+
+**Method signature:** 
+
+```
+function toggleUnionBridgeIntegrationPause() public returns (int256);
+```
```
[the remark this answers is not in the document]
marcos-iov: Changed to a single method that allows RBTC transfers between the PowPeg and the Union Bridge to be independently enabled or disabled.

IPs/RSKIP502.md
```diff
@@ -0,0 +1,264 @@
...
+
+Special considerations have been made to allow seamless testing and updates during the Union Bridge's development phase, with restrictions placed on mainnet interactions to avoid unauthorized changes.
+
+## Motivation
+
+The Union Bridge is a new project currently under development that will interact with the Rootstock network. Integrating the Union Bridge with the existing PowPeg is essential to facilitate the transfer of RBTC. A key concern is ensuring that the integration remains secure, and that risks are minimized through a locking cap that controls the amount of RBTC transferred. The PowPeg's current functionality must be extended with minimal disruption, while also accounting for the need to manage the evolving nature of the Union Bridge.
+
+Additionally, the Union Bridge is still in development, and its contract address is likely to change as it evolves. To accommodate this, the proposed solution includes a flexible mechanism for updating the Union Bridge contract address during its testing and development phases, ensuring that these changes can occur securely without risking mainnet operations.
```
nathanieliov, line 35: I would suggest doing some rewording to this paragraph, the last part, "ensuring that these changes can occur securely without risking mainnet operations" gives me the impression that this functionality to update the union bridge contract address will be available to the mainnet network too.

aeidelman (approved):

PR #517, opened 2025-05-29
RSKIP517: Block time-centric difficulty adjustment with uncle threshold

patogallaiovlabs:
This RSKIP proposes a change of the difficulty adjustment algorithm to primarily consider block time over uncle rate. The primary goal is to reduce the network's average block time by lowering the mining difficulty. However, reducing the difficulty increases the risk of uncle block generation due to current mining pool behaviors related to block template refresh intervals. To address this, the proposal introduces a threshold-based mechanism that allows difficulty to decrease without being overly penalized by temporary increases in uncle rate, thereby improving network stability. The current method also relies on per-block signals, which often misrepresent broader network trends due to the statistical nature of block time distribution. This proposal introduces a window-based averaging mechanism that better estimates the underlying block time. This approach yields a more stable and accurate approximation of network performance.


smishraIOV (approved):
The proposal is clear and easy to follow. Currently, we update difficulty every block - which is perhaps a legacy from ethereum's early design. You propose a difficulty update every 30 blocks. In next version, explain this a bit more

IPs/RSKIP517.md
```diff
@@ -0,0 +1,144 @@
...
+The new difficulty adjustment introduces a conditional factor $F$ that adapts based on both the uncle rate and deviation from the target block time. This mechanism ensures that uncle blocks impact difficulty when their rate exceeds a defined threshold, preserving network stability while allowing controlled block time reductions. The adjustment factor is defined as:
+
+$$
+F = \begin{cases}
+\alpha & \text{if } R > C \\
+\alpha & \text{if } R \leq C \land A > T \cdot 1.10 \\
+-\alpha & \text{if } R  \leq C  \land A < T \cdot 0.90 \\
+0 & \text{otherwise}
+\end{cases}
+$$
```
SergioDemianLerner, line 101: The actual computation should be specified because depending on how this is computed (bigint arithmetic, float arithmetic, etc.) it may give different results.

I suggest to add. here the exact variable types that will be used and analyze the performance of those computations. Maybe computing the coheficients using fixed-point numbers (i.e. 2<<10 multiplier) in uint64 variables can be more efficient.

PR #546, opened 2026-01-27
rskip to introduce ethereum's type 1 and 2 transaction format

smishraIOV:
First draft of proposal to implement new eip-2718 transaction type envelope encoding for type 1 and type 2 txs. 
* The proposal is for encoding only: access lists and eip-1559 gas pricing will NOT be implemented
* the new formats will be executed just like legacy txs, with additional gas cost for including any data in the "accessList" field
* This can help with Account Abstraction EIP-7702


IPs/RSKIP546.md
```diff
@@ -0,0 +1,106 @@
...
+The `signatureYParity`, `signatureR`, `signatureS` elements of this transaction represent a secp256k1 signature over `keccak256(0x01 || rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, accessList]))`.
+
+
+After validating the signature, and ignoring the `accessList` field, the transaction can be processed in the same way as a legacy transaction. See the "Transaction Normalization and Additional Gas" section below.
+
+The EIP-2718 `ReceiptPayload` for this transaction is `rlp([status, cumulativeGasUsed, logsBloom, logs])`. Thus, the full receipt will be `01 || ReceiptPayload`.
+
+### Type 2 Transactions
+
+The EIP-2718 TransactionPayload for this transaction type is `rlp([chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, amount, data, access_list, signature_y_parity, signature_r, signature_s])`.
```
meldsun0, line 55: Type 1 transaction is described as ... signatureYParity, signatureR, signatureS]), whereas Type 2 transaction uses the same structure but with different naming: signature_y_parity, signature_r, signature_s]).

smishraIOV: Thank you for pointing out the inconsistency! Do you recommend that we make them the same? Or should we add a clarifying note that the inconsistency is because we are following the notation from original EIPs?


Legacy:  `v, r, s` :   https://eips.ethereum.org/EIPS/eip-2718
type-1:  `signatureYParity, signatureR, signatureS` : https://eips.ethereum.org/EIPS/eip-2930
type-2 : `signature_y_parity, signature_r, signature_s` : https://eips.ethereum.org/EIPS/eip-1559
type-3:  `signature_y_parity, signature_r, signature_s` : https://eips.ethereum.org/EIPS/eip-4844
type-4:  `y_parity, r, s` : https://eips.ethereum.org/EIPS/eip-7702

smishraIOV: I updated it to clarify why the inconsistency exists.

aeidelman (approved):

PR #555, opened 2026-04-03
RSKIP555: Update gas cost calculation to match Ethereum Berlin hardfork

wlaclaw:


IPs/RSKIP555.md
```diff
@@ -0,0 +1,460 @@
...
+
+### 7. `CREATE` (0xF0) and `CREATE2` (0xF5)
+
+The address of the newly created contract is inserted into `accessed_addresses`. No access charge is made for the insertion.
+
+### 8. `SSTORE` (0x55)
+
+`SSTORE` pricing changes in two parts, both applied at the same activation height.
+
+**First, the cold surcharge.** Let `key` be the storage key and `addr` the executing contract. If `(addr, key)` isn't in `accessed_storage_keys`, charge `COLD_SLOAD_COST` and insert it. This charge is made *in addition to* the cost computed below.
```
italo-sampaio, line 161: The order here reads as "charge the cold surcharge, then run the EIP-2200 algorithm", and step 1 of that algorithm is the EIP-1706 sentry. Taken literally, the sentry sees the gas left after the 2100 has already been taken. Ethereum clients don't do that.

Example: a frame reaches `SSTORE` with 4400 gas on a cold slot where `current == new`. Under the literal order it pays 2100, has 2300 left and fails the sentry. In geth and the execution-specs, the sentry reads gas before any part of the opcode's cost. The same frame passes, pays 2100 + 100 and continues with 2200. Every cold `SSTORE` entered with gas in (2300, 4400] diverges.

- geth, `makeGasSStoreFunc` in [`core/vm/operations_acl.go`](https://github.com/ethereum/go-ethereum/blob/master/core/vm/operations_acl.go): the `SstoreSentryGasEIP2200` check runs before the access-list lookup, and the cold cost is added to a single charge.
- execution-specs, Berlin `sstore`: `gas_left <= CALL_STIPEND` is checked before the cold cost is added.

Suggestion: state the order explicitly. The sentry compares against the gas remaining before this opcode's cost. Then the EIP-2200 cost is computed, the cold surcharge is added and the key inserted, and the total is charged once. Test case 19 has the same ambiguity ("2300 or less gas remaining"). A cold-slot vector at 4400 gas would pin it down.

IPs/RSKIP555.md
```diff
@@ -0,0 +1,460 @@
...
+        - If `new` is 0, add `SSTORE_CLEARS_SCHEDULE` (15000) to the refund counter.
+      - If `original` equals `new`, so the slot has been reset to where it started:
+        - If `original` is 0, add `SSTORE_SET_GAS - SLOAD_GAS` (19900) to the refund counter.
+        - Otherwise, add `SSTORE_RESET_GAS - SLOAD_GAS` (2800) to the refund counter.
+
+Refund policy is otherwise unchanged. `SSTORE_CLEARS_SCHEDULE` stays at 15000 and the `SELFDESTRUCT` refund is retained. Ethereum's later reform of both ([EIP-3529][eip3529]) isn't adopted here.
+
+### 9. Reversion
+
+If a call frame reverts, `accessed_addresses` and `accessed_storage_keys` are restored to the contents they had when that frame began. Entries added by a frame that reverts don't stay warm for later frames.
```
italo-sampaio, line 193: This only covers a frame that reverts. A frame can also end in an exceptional halt: out-of-gas, an invalid opcode, a stack violation, or a state change under `STATICCALL`. EIP-2929 anchors this by saying the sets are "implemented identically to other transaction-scoped constructs such as the self-destruct-list and global `refund` counter". Those constructs are discarded on any failed frame. That sentence was dropped here, so a REVERT-only reading is possible.

Under that reading, an address warmed in a child frame that runs out of gas would stay warm in the caller. The next access would cost 100 instead of 2600. That includes running out of gas while paying the cold cost itself: geth inserts before charging and relies on the frame rollback to undo it. rskj already treats every other piece of frame state this way. `Program` rolls back when `exception != null || isRevert()`, and `ProgramResult.merge` only folds a child's effects upward on success.

Suggestion: say "If a call frame reverts or halts exceptionally", and restore EIP-2929's sentence tying the access sets to the self-destruct list and the refund counter. Test cases 4 and 5 could each get an out-of-gas variant.

IPs/RSKIP555.md
```diff
@@ -0,0 +1,460 @@
...
+
+### 6. `SELFDESTRUCT` (0xFF)
+
+If the beneficiary address isn't in `accessed_addresses`, charge `COLD_ACCOUNT_ACCESS_COST` **in addition to** the existing cost, and insert it. If it's already present, no additional charge is made.
+
+The existing `SUICIDE` (5000) and `NEW_ACCT_SUICIDE` (25000) costs and the `SUICIDE_REFUND` (24000) refund are unchanged.
+
+### 7. `CREATE` (0xF0) and `CREATE2` (0xF5)
+
+The address of the newly created contract is inserted into `accessed_addresses`. No access charge is made for the insertion.
```
italo-sampaio, line 155: This doesn't say when the address is inserted, or whether it stays warm if the creation fails. "Insert only on success" and "insert before running the initcode" are both valid interpretations. They differ by 2500 gas on the next access to that address after an initcode revert, an out-of-gas or an address collision.

EIP-2929 says the address is added "immediately (before checks are done to determine whether or not the address is unclaimed)" and that it "remains in `accessed_addresses`" if the creation later fails. In practice, clients insert after the depth, balance and nonce pre-checks and before the collision check:

- geth: `create` in [`core/vm/evm.go`](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go) runs the pre-checks first, then adds the address to the access list "_before_ taking a snapshot".
- execution-specs, Berlin: `generic_create` returns on a balance, nonce or depth failure before `accessed_addresses.add`, which comes before the `account_deployable` collision check.

Copying "immediately" alone would therefore warm the address in cases where Ethereum doesn't.

Suggestion: say the address is inserted once the depth, balance and nonce checks have passed and before the collision check, and that it stays warm if the creation fails afterwards. The insertion belongs to the creating frame, so §9 only removes it if that frame itself fails. A failed-create variant of test case 16 would pin this down.

IPs/RSKIP555.md
```diff
@@ -0,0 +1,460 @@
...
+
+1. If remaining gas is less than or equal to `SSTORE_SENTRY_GAS` (2300), fail the current call frame with an out-of-gas exception. This is [EIP-1706][eip1706], a component of EIP-2200.
+2. If `current` equals `new`, charge `SLOAD_GAS` (100).
+3. If `current` doesn't equal `new`:
+   1. If `original` equals `current`, so the slot hasn't been modified yet in this transaction:
+      - If `original` is 0, charge `SSTORE_SET_GAS` (20000).
+      - Otherwise, charge `SSTORE_RESET_GAS` (2900). If `new` is 0, add `SSTORE_CLEARS_SCHEDULE` (15000) to the refund counter.
+   2. If `original` doesn't equal `current`, so the slot is dirty, charge `SLOAD_GAS` (100), then apply both of the following:
+      - If `original` isn't 0:
+        - If `current` is 0, subtract `SSTORE_CLEARS_SCHEDULE` (15000) from the refund counter.
```
italo-sampaio, line 183: This branch can take a frame's refund counter below zero. Example: a slot originally holds X. The parent writes 0 and gets +15000. A reentrant child then writes Y, which lands here and subtracts 15000. On Ethereum the net is zero.

In rskj, the refund is accumulated per frame and merged into the parent only on success. `GasCost.add` and `GasCost.subtract` throw on negative values. A direct port therefore either throws, failing the child frame where Ethereum succeeds, or clamps at zero and over-refunds 15000. Both diverge from Ethereum.

EIP-2200 already addresses this: "if the implementation uses call-frame refund counter, the counter can go negative. If the implementation uses transaction-wise refund counter, the counter always stays positive."

Suggestion: carry that sentence into §8. The per-frame counter is then explicitly signed, and only the transaction total is guaranteed non-negative when the existing `gasUsed / 2` cap is applied.

italo-sampaio (commented):
left a few comments inline

PR #558, opened 2026-04-10
Introduce Blob Txs (eip-4844)

smishraIOV:
This proposal is to implement type-3 (blob) transactions which can help L2 rollups with data availability. Following up on RSKIPs 543, 545 and 546 for transaction types 1, 2, and 4.


Nuhvi:
I would like to voice my support and want of ephemeral data on Rootstock, for the purpose of proof of publication. Plenty can be done with a decentralized censorship resistant notary, even without any smart contracts. For many applications, the entire logic can be client side validated, and the computation compressed as a STARK proof.

These applications can still be done on Rootstock today, but it would either bloat the state, or bloat the history or both. History bloat might not be an issue, given that there are other ways to make sidechains ephemeral even without Blobs. But the pricing will still enable doing more, if nodes know they don't have to store the data forever.

patogallaiovlabs (approved):

patogallaiovlabs (approved):

PR #575, opened 2026-07-29
RSKIP-169: set status to Adopted

ilanolkies:
Align the preamble and header-table status (Draft) with the canonical index (Adopted); the EXTCODEHASH fix is implemented and gated behind ConsensusRule.RSKIP169.


raullaprida (approved):
lgtm

PR #577, opened 2026-07-30
RSKIP-16 (Combined State Tree): set status to Adopted

ilanolkies:
Aligns the recorded status (frontmatter, body table, README index) with what the code shows: rskj ships the Unitrie this RSKIP proposes — a single trie combining accounts, contract storage and code (MutableRepository over one Trie, the `unitrie_` store prefix). Confidence is MEDIUM: this is the foundational-idea case (later RSKIPs formalized it) rather than a single gated flag, so worth a second look.

### DIFF
--- a/IPs/RSKIP16.md
+++ b/IPs/RSKIP16.md
@@ -2,7 +2,7 @@
 rskip: 16
 title: Combined State Tree 
 description: 
-status: Draft
+status: Adopted
 purpose: Sca
 author: SDL (@sergiodemianlerner)
 layer: Core
@@ -20,7 +20,7 @@ created: 2016-11-01
 |**Purpose**    |Sca |
 |**Layer**      |Core |
 |**Complexity** |3 |
-|**Status**     |Draft |
+|**Status**     |Adopted |
 
 ## Pre-git revisions
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ You can find an easily browseable version of this information [here](https://ips
 | 13       | [Support for stable assets & token issuance](IPs/RSKIP13.md)                     | 08-AUG-16 | SDL       | Sca      | Core     | 3 | Draft    |
 | 14       | [Reward Manager Smart Contract (REMASC)](IPs/RSKIP14.md)                         | 10-NOV-16 | SDL       | Sca      | Core     | 3 | Rejected |
 | 15       | [Simplified Reward Manager Smart Contract (REMASC)](IPs/RSKIP15.md)              | 14-NOV-16 | SDL       | Sca      | Core     | 3 | Adopted  |
-| 16       | [Combined State Tree](IPs/RSKIP16.md)                                            | 01-NOV-16 | SDL       | Sca      | Core     | 3 | Draft    |
+| 16       | [Combined State Tree](IPs/RSKIP16.md)                                            | 01-NOV-16 | SDL       | Sca      | Core     | 3 | Adopted  |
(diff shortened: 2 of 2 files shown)


ilanolkies:
I did a field-by-field check of this RSKIP's Specification section against rskj (`e7de490`) before confirming the status flip. The structural spec is implemented literally in [`TrieKeyMapper`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java):

| RSKIP-16 says | rskj has | Match |
|---|---|---|
| Trie key prefixed by namespace byte `0x00` | [`DOMAIN_PREFIX = {0x00}`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L42) | ✅ exact |
| Account field selector `0x00` = storage cells | [`STORAGE_PREFIX = {0x00}`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L43) | ✅ exact |
| Account field selector `0x80` = code | [`CODE_PREFIX = {0x80}`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L44) | ✅ exact |
| Path: namespace → 20-byte account address → selector → storage address | [`mapRskAddressToKey`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L89-L92) / [`getAccountStorageKey`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L69-L75) build exactly this shape | ✅ |
| Single tree for accounts + storage + code | [`MutableRepository`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/MutableRepository.java) resolves account state, code and storage cells through these keys over one `Trie`; `AccountState` holds only nonce+balance since code/storage became sub-tree nodes | ✅ |

Two details deviate from this document's letter, and both are exactly what RSKIP-108 later refined: keys are not fully hash-randomized (sha3-replaced) as written here — the code uses a 10-byte keccak prefix plus the plain key ([`secureKeyPrefix`](https://github.com/rsksmart/rskj/blob/e7de490f2a487f99e0384fd6f729881ca99893d4/rskj-core/src/main/java/org/ethereum/db/TrieKeyMapper.java#L77-L79)) — and storage keys are leading-zero-stripped rather than fixed 32 bytes.

So: the proposal's structure (namespace byte, account sub-tree, `0x00`/`0x80` selectors, unified tree) is what mainnet ships, with the key-hashing details superseded by RSKIP-108. On that basis Adopted looks right for this document; the alternative reading would be to prefer marking it superseded by RSKIP-108/107 if we want status to track the letter rather than the design. Flagging both so the review can settle it.

> 108 doesn't suppress 16, it amends one detail of it. RSKIP-16 defines the structure (single tree, 0x00 namespace, account sub-tree, 0x00/0x80 selectors) and all of that ships literally. RSKIP-108 only replaces the key-hashing rule inside that structure (80-bit hash prefix + plain key instead of full sha3 replacement, zero-stripped storage keys). They compose; both can legitimately be Adopted.

PR #583, opened 2026-08-03
RSKIP-32 (Double-Hashed Addresses): set status to Draft

ilanolkies:
Aligns the README index row for RSKIP-32 (Double-Hashed Addresses) with the spec's recorded status: `Draft*` becomes plain `Draft`. The proposal was never implemented — rskj derives addresses with the plain single-hash 20-byte scheme (RskAddress, AccountState), with no double-hash derivation anywhere in rskj-core — so Draft is the correct status.


ilanolkies:
Superseded: RSKIP-32's recorded status changed to Withdrawn — a replacement PR follows.

PR #678, opened 2026-08-27
Add RSKIP678 placeholder: Next release improvements

fmacleal:
Adding a placeholder RSKIP to hold some improvements planned for next release that require hardfork.


Nuhvi:
If the plans for next release isn't in RSKIP nor on research.rsk.dev, then where is it and where is it usually discussed?

